PR #2, opened 2016-05-09
everything

davidchambers:
This pull request defines the following:

  - `Z.TypeClass :: (String, Array TypeClass, a -> Boolean) -> TypeClass`
  - `Z.Setoid :: TypeClass`
  - `Z.Semigroup :: TypeClass`
  - `Z.Monoid :: TypeClass`
  - `Z.Functor :: TypeClass`
  - `Z.Bifunctor :: TypeClass`
  - `Z.Profunctor :: TypeClass`
  - `Z.Apply :: TypeClass`
  - `Z.Applicative :: TypeClass`
  - `Z.Chain :: TypeClass`
  - `Z.ChainRec :: TypeClass`
  - `Z.Monad :: TypeClass`
  - `Z.Foldable :: TypeClass`
  - `Z.Traversable :: TypeClass`
  - `Z.Extend :: TypeClass`
  - `Z.Comonad :: TypeClass`
  - `Z.toString :: Showable a => a -> String`
  - `Z.equals :: (a, b) -> Boolean`
  - `Z.concat :: Semigroup a => (a, a) -> a`
  - `Z.empty :: Monoid m => TypeRep m -> m`
  - `Z.map :: Functor f => (a -> b, f a) -> f b`
  - `Z.bimap :: Bifunctor f => (a -> b, c -> d, f a c) -> f b d`
  - `Z.promap :: Profunctor p => (a -> b, c -> d, p b c) -> p a d`
  - `Z.ap :: Apply f => (f (a -> b), f a) -> f b`
  - `Z.of :: Applicative f => (TypeRep f, a) -> f a`
  - `Z.chain :: Chain m => (a -> m b, m a) -> m b`
  - `Z.chainRec :: ChainRec m => (TypeRep m, (a -> c, b -> c, a) -> m c, a) -> m b`
  - `Z.filter :: (Applicative f, Foldable f, Monoid (f a)) => (a -> Boolean, f a) -> f a`
  - `Z.filterM :: (Monad m, Monoid (m a)) => (a -> Boolean, m a) -> m a`
  - `Z.reduce :: Foldable f => ((b, a) -> b, b, f a) -> b`
  - `Z.traverse :: (Applicative f, Traversable t) => (a -> f a, b -> f c, t b) -> f (t c)`
  - `Z.sequence :: (Applicative f, Traversable t) => (a -> f a, t (f b)) -> f (t b)`
  - `Z.extend :: Extend w => (w a -> b, w a) -> w b`
  - `Z.extract :: Comonad w => w a -> a`

---

Original description:

> This pull request is not yet ready to merge, but it's *well* past my bedtime!
>
> In sanctuary-js/sanctuary-def#62 @kedashoe suggested creating a repository for type classes. Having all the Fantasy Land type classes neatly defined in one place sounded great to me, so I've since spent some time making it happen.
>
> While writing the predicates I realized we want Array to be considered a Chain (among other type classes) but `Array#chain` is not defined. This presents us with three options:
>
>   - force Sanctuary to provide its *own* Chain type class when we define `S.chain`;
>
>   - claim that Array satisfies the requirements of Chain and hope that somehow it does (because libraries such as Sanctuary add logic to handle the missing method); or
>
>   - define `chain` for Array, and provide a `chain` function which makes use of this definition.
>
> The last option seems best to me. We'll have all the definitions of Fantasy Land methods for built-in types in one place rather than scattered through a dozen or so files in Ramda. :)
>
> Note that while the methods are in place, I've yet to define the functions to wrap them. We should avoid invoking methods directly so we don't invoke missing methods such as `Array#chain`, and to ensure we use our lawful `map` rather than `Array#map`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,81 @@
+DOCTEST = node_modules/.bin/doctest --nodejs '--harmony' --module commonjs --prefix .
+ESLINT = node_modules/.bin/eslint --config node_modules/sanctuary-style/eslint-es3.json --env es3
+ISTANBUL = node_modules/.bin/istanbul
+NPM = npm
+PREDOCTEST = scripts/predoctest
+TRANSCRIBE = node_modules/.bin/transcribe
+XYZ = node_modules/.bin/xyz --repo git@github.com:sanctuary-js/sanctuary-type-classes.git --script scripts/prepublish
+
+TEST = $(shell find test -name '*.js' | sort)
+
+
+.PHONY: all
+all: LICENSE README.md
+
+.PHONY: LICENSE
+LICENSE:
+	cp -- '$@' '$@.orig'
+	sed 's/Copyright (c) .* Sanctuary/Copyright (c) $(shell git log --date=format:%Y --pretty=format:%ad | sort -r | head -n 1) Sanctuary/' '$@.orig' >'$@'
+	rm -- '$@.orig'
+
+README.md: index.js.tmp
+	$(TRANSCRIBE) \
+	  --heading-level 4 \
+	  --url 'https://github.com/sanctuary-js/sanctuary-type-classes/blob/v$(VERSION)/index.js#L{line}' \
+	  -- '$<' \
+	| LC_ALL=C sed 's/<h4 name="\(.*\)#\(.*\)">\(.*\)\1#\2/<h4 name="\1.prototype.\2">\3\1#\2/' >'$@'
+
+.INTERMEDIATE: index.js.tmp
+index.js.tmp: index.js
+	sed -e '/^[/][/]:/ s,\([[:<:]][[:alnum:]]*\),<a href="#\1">\1</a>,g' -e 's,^//:,//.,' '$<' >'$@'
+
+
+.PHONY: doctest
+doctest: index-no-blockquotes.js
+ifeq ($(shell node --version | cut -d . -f 1),v6)
+	$(DOCTEST) -- $^
+else
+	@echo '[WARN] Doctests are only run in Node v6.x.x (current version is $(shell node --version))' >&2
+endif
+
+.INTERMEDIATE: index-no-blockquotes.js
+index-no-blockquotes.js: index.js $(PREDOCTEST)
+	$(PREDOCTEST) '$<' >'$@'
+
+
+.PHONY: lint
+lint:
+	$(ESLINT) \
+	  --global define \
+	  --global module \
+	  --global self \
+	  --rule 'max-len: [error, {code: 79, ignoreUrls: true, ignorePattern: "^ *//(# |  .* :: |[.] > )"}]' \
+	  --rule 'spaced-comment: [error, always, {line: {exceptions: ["."], markers: ["#", ".", ":"]}}]' \
+	  -- index.js
+	$(ESLINT) \
+	  --env node \
+	  -- $(PREDOCTEST) $(TEST)
+	@echo 'Checking for missing link definitions...'
+	grep -o '\[[^]]*\]\[[^]]*\]' index.js \
+	| sort -u \
+	| sed -e 's:\[\(.*\)\]\[\]:\1:' \
+	      -e 's:\[.*\]\[\(.*\)\]:\1:' \
+	      -e '/0-9/d' \
+	| xargs -I '{}' sh -c "grep '^//[.] \[{}\]: ' index.js"
+
+
+.PHONY: release-major release-minor release-patch
+release-major release-minor release-patch:
+	@$(XYZ) --increment $(@:release-%=%)
+
+
+.PHONY: setup
+setup:
+	$(NPM) install
+
+
+.PHONY: test
+test:
+	$(ISTANBUL) cover node_modules/.bin/_mocha -- --timeout 10000 --ui tdd -- test/index.js
+	$(ISTANBUL) check-coverage --branches 100
+	make doctest
--- a/bower.json
+++ b/bower.json
@@ -1,0 +1,35 @@
+{
+  "name": "sanctuary-type-classes",
+  "description": "Standard library for Fantasy Land",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/sanctuary-js/sanctuary-type-classes.git"
+  },
+  "main": "index.js",
+  "moduleType": [
+    "amd",
+    "globals",
+    "node"
+  ],
+  "keywords": [
+    "algebraic",
+    "functional",
+    "sanctuary",
+    "types"
+  ],
+  "dependencies": {},
+  "ignore": [
+    "/.git/",
+    "/bower_components/",
+    "/coverage/",
+    "/node_modules/",
+    "/scripts/",
+    "/test/",
+    "/.*",
+    "/CONTRIBUTING.md",
+    "/Makefile",
+    "/circle.yml",
+    "/package.json"
+  ]
+}
--- a/index.js
+++ b/index.js
@@ -1,0 +1,1299 @@
+/*
+             ############                  #
+            ############                  ###
+                  #####                  #####
+                #####      ####################
+              #####       ######################
+            #####                     ###########
+          #####         ######################
+        #####          ####################
+      #####                        #####
+     ############                 ###
+    ############                 */
+
+//. # sanctuary-type-classes
+//.
+//. The [Fantasy Land Specification][FL] "specifies interoperability of common
+//. algebraic structures" by defining a number of type classes. For each type
+//. class, it states laws which every member of a type must obey in order for
+//. the type to be a member of the type class. In order for the Maybe type to
+//. be considered a [Functor][], for example, every `Maybe a` value must have
+//. a `fantasy-land/map` method which obeys the identity and composition laws.
+//.
+//. This project provides:
+//.
+//.   - [`TypeClass`](#TypeClass), a function for defining type classes;
+//.   - one `TypeClass` value for each Fantasy Land type class;
+//.   - lawful Fantasy Land methods for JavaScript's built-in types;
+//.   - one function for each Fantasy Land method; and
+//.   - several functions derived from these functions.
+//.
+//. ## Type-class hierarchy
+//.
+//. <pre>
+//:  Setoid   Semigroup   Foldable           Functor            Extend
+//: (equals)   (concat)   (reduce)            (map)            (extend)
+//:               |           \             / | | | \             /
+//:               |            \           /  | | |  \           /
+//:               |             \         /   | | |   \         /
+//:               |              \       /    | | |    \       /
+//:               |               \     /     | | |     \     /
+//:               |                \   /      | | |      \   /
+//:               |                 \ /       / | \       \ /
+//:            Monoid           Traversable  /  |  \    Comonad
+//:            (empty)          (traverse)  /   |   \  (extract)
+//:                                        /    |    \
+//:                               Bifunctor   Apply   Profunctor
+//:                                (bimap)     (ap)    (promap)
+//:                                            / \
+//:                                           /   \
+//:                                          /     \
+//:                                         /       \
+//:                                        /         \
+//:                                  Applicative    Chain
+//:                                      (of)      (chain)
+//:                                        \         / \
+//:                                         \       /   \
+//:                                          \     /     \
+//:                                           \   /       \
+//:                                            \ /         \
+//:                                           Monad     ChainRec
+//:                                                    (chainRec)
+//. </pre>
+//.
+//. ## API
+
+(function(f) {
+
+  'use strict';
+
+  /* istanbul ignore else */
+  if (typeof module === 'object' && typeof module.exports === 'object') {
+    module.exports = f();
+  } else if (typeof define === 'function' && define.amd != null) {
+    define([], f);
+  } else {
+    self.sanctuaryTypeClasses = f();
+  }
+
+}(function() {
+
+  'use strict';
+
+  //  has :: (String, Object) -> Boolean
+  var has = function(k, o) {
+    return Object.prototype.hasOwnProperty.call(o, k);
+  };
+
+  //  identity :: a -> a
+  var identity = function(x) { return x; };
+
+  //  prepend :: a -> Array a -> Array a
+  var prepend = function(x) {
+    return function(xs) {
+      return [x].concat(xs);
+    };
+  };
+
+  //  type :: Any -> String
+  var type = function type(x) {
+    return x != null && type(x['@@type']) === 'String' ?
+      x['@@type'] :
+      Object.prototype.toString.call(x).slice('[object '.length, -']'.length);
+  };
+
+  //# TypeClass :: (String, Array TypeClass, a -> Boolean) -> TypeClass
+  //.
+  //. The arguments are:
+  //.
+  //.   - the name of the type class, prefixed by its npm package name;
+  //.   - an array of dependencies; and
+  //.   - a predicate which accepts any JavaScript value and returns `true`
+  //.     if the value satisfies the requirements of the type class; `false`
+  //.     otherwise.
+  //.
+  //. Example:
+  //.
+  //. ```javascript
+  //. //    hasMethod :: String -> a -> Boolean
+  //. const hasMethod = name => x => x != null && typeof x[name] == 'function';
+  //.
+  //. //    Foo :: TypeClass
+  //. const Foo = Z.TypeClass('my-package/Foo', [], hasMethod('foo'));
+  //.
+  //. //    Bar :: TypeClass
+  //. const Bar = Z.TypeClass('my-package/Bar', [Foo], hasMethod('bar'));
+  //. ```
+  //.
+  //. Types whose values have a `foo` method are members of the Foo type class.
+  //. Members of the Foo type class whose values have a `bar` method are also
+  //. members of the Bar type class.
+  //.
+  //. Each `TypeClass` value has a `test` field: a function which accepts
+  //. any JavaScript value and returns `true` if the value satisfies the
+  //. type class's predicate and the predicates of all the type class's
+  //. dependencies; `false` otherwise.
+  //.
+  //. `TypeClass` values may be used with [sanctuary-def][type-classes]
+  //. to define parametrically polymorphic functions which verify their
+  //. type-class constraints at run time.
+  var TypeClass = function(name, dependencies, test) {
+    return {
+      '@@type': 'sanctuary-type-classes/TypeClass',
+      name: name,
+      test: function(x) {
+        return dependencies.every(function(d) { return d.test(x); }) &&
+               test(x);
+      }
+    };
+  };
+
+  //  orConstructor :: (String, Object) -> Function?
+  var orConstructor = function(methodName, x) {
+    return typeof x[methodName] === 'function' ? x : x.constructor;
+  };
+
+  //  data Location = Constructor | Value
+
+  //  Constructor :: Location
+  var Constructor = 'Constructor';
+
+  //  Value :: Location
+  var Value = 'Value';
+
+  //  _funcPath :: (Boolean, Array String, a) -> Nullable Function
+  var _funcPath = function(allowInheritedProps, path, _x) {
+    var x = _x;
+    for (var idx = 0; idx < path.length; idx += 1) {
+      var k = path[idx];
+      if (x == null || !(allowInheritedProps || has(k, x))) return null;
+      x = x[k];
+    }
+    return typeof x === 'function' ? x : null;
+  };
+
+  //  funcPath :: (Array String, a) -> Nullable Function
+  var funcPath = function(path, x) {
+    return _funcPath(true, path, x);
+  };
+
+  //  implPath :: Array String -> Nullable Function
+  var implPath = function(path) {
+    return _funcPath(false, path, implementations);
+  };
+
+  //  $ :: (String, Array TypeClass, StrMap (Array Location)) -> TypeClass
+  var $ = function(_name, dependencies, requirements) {
+    var getBoundMethod = function(_name) {
+      return function(x) {
+        var name = 'fantasy-land/' + _name;
+        var locations = requirements[_name];
+        for (var idx = 0; idx < locations.length; idx += 1) {
+          if (locations[idx] === Constructor) {
+            var f = funcPath(['constructor', name], x) ||
+                    implPath([type(x), name]);
+            /* istanbul ignore else */  // Constructor always listed last
+            if (f) return f;
+          } else {
+            var m = funcPath([name], x) ||
+                    implPath([type(x), 'prototype', name]);
+            if (m) return m.bind(x);
+          }
+        }
+      };
+    };
+
+    var name = 'sanctuary-type-classes/' + _name;
+    var keys = Object.keys(requirements);
+
+    var typeClass = TypeClass(name, dependencies, function(x) {
+      return keys.every(function(_name) {
+        return getBoundMethod(_name)(x) != null;
+      });
+    });
+
+    typeClass.functions = {};
+    keys.forEach(function(_name) {
+      var name = 'fantasy-land/' + _name;
+      if (requirements[_name].indexOf(Constructor) >= 0) {
+        typeClass.functions[_name] = function(typeRep) {
+          var m = /function (\w*)/.exec(typeRep);
+          return m && implPath([m[1], name]) || typeRep[name];
+        };
+      }
+    });
+
+    typeClass.methods = {};
+    keys.forEach(function(_name) {
+      /* istanbul ignore else */  // Value always listed
+      if (requirements[_name].indexOf(Value) >= 0) {
+        typeClass.methods[_name] = getBoundMethod(_name);
+      }
+    });
+
+    return typeClass;
+  };
+
+  //# Setoid :: TypeClass
+  //.
+  //. `TypeClass` value for [Setoid][].
+  //.
+  //. ```javascript
+  //. > Setoid.test(null)
+  //. true
+  //. ```
+  var Setoid = $('Setoid', [], {equals: [Value]});
+
+  //# Semigroup :: TypeClass
+  //.
+  //. `TypeClass` value for [Semigroup][].
+  //.
+  //. ```javascript
+  //. > Semigroup.test('')
+  //. true
+  //.
+  //. > Semigroup.test(0)
+  //. false
+  //. ```
+  var Semigroup = $('Semigroup', [], {concat: [Value]});
+
+  //# Monoid :: TypeClass
+  //.
+  //. `TypeClass` value for [Monoid][].
+  //.
+  //. ```javascript
+  //. > Monoid.test('')
+  //. true
+  //.
+  //. > Monoid.test(0)
+  //. false
+  //. ```
+  var Monoid = $('Monoid', [Semigroup], {empty: [Value, Constructor]});
+
+  //# Functor :: TypeClass
+  //.
+  //. `TypeClass` value for [Functor][].
+  //.
+  //. ```javascript
+  //. > Functor.test([])
+  //. true
+  //.
+  //. > Functor.test('')
+  //. false
+  //. ```
+  var Functor = $('Functor', [], {map: [Value]});
+
+  //# Bifunctor :: TypeClass
+  //.
+  //. `TypeClass` value for [Bifunctor][].
+  //.
+  //. ```javascript
+  //. > Bifunctor.test(Tuple('foo', 64))
+  //. true
+  //.
+  //. > Bifunctor.test([])
+  //. false
+  //. ```
+  var Bifunctor = $('Bifunctor', [Functor], {bimap: [Value]});
+
+  //# Profunctor :: TypeClass
+  //.
+  //. `TypeClass` value for [Profunctor][].
+  //.
+  //. ```javascript
+  //. > Profunctor.test(Math.sqrt)
+  //. true
+  //.
+  //. > Profunctor.test([])
+  //. false
+  //. ```
+  var Profunctor = $('Profunctor', [Functor], {promap: [Value]});
+
+  //# Apply :: TypeClass
+  //.
+  //. `TypeClass` value for [Apply][].
+  //.
+  //. ```javascript
+  //. > Apply.test([])
+  //. true
+  //.
+  //. > Apply.test({})
+  //. false
+  //. ```
+  var Apply = $('Apply', [Functor], {ap: [Value]});
+
+  //# Applicative :: TypeClass
+  //.
+  //. `TypeClass` value for [Applicative][].
+  //.
+  //. ```javascript
+  //. > Applicative.test([])
+  //. true
+  //.
+  //. > Applicative.test({})
+  //. false
+  //. ```
+  var Applicative = $('Applicative', [Apply], {of: [Value, Constructor]});
+
+  //# Chain :: TypeClass
+  //.
+  //. `TypeClass` value for [Chain][].
+  //.
+  //. ```javascript
+  //. > Chain.test([])
+  //. true
+  //.
+  //. > Chain.test({})
+  //. false
+  //. ```
+  var Chain = $('Chain', [Apply], {chain: [Value]});
+
+  //# ChainRec :: TypeClass
+  //.
+  //. `TypeClass` value for [ChainRec][].
+  //.
+  //. ```javascript
+  //. > ChainRec.test([])
+  //. true
+  //.
+  //. > ChainRec.test({})
+  //. false
+  //. ```
+  var ChainRec = $('ChainRec', [Chain], {chainRec: [Value, Constructor]});
+
+  //# Monad :: TypeClass
+  //.
+  //. `TypeClass` value for [Monad][].
+  //.
+  //. ```javascript
+  //. > Monad.test([])
+  //. true
+  //.
+  //. > Monad.test({})
+  //. false
+  //. ```
+  var Monad = $('Monad', [Applicative, Chain], {});
+
+  //# Foldable :: TypeClass
+  //.
+  //. `TypeClass` value for [Foldable][].
+  //.
+  //. ```javascript
+  //. > Foldable.test({})
+  //. true
+  //.
+  //. > Foldable.test('')
+  //. false
+  //. ```
+  var Foldable = $('Foldable', [], {reduce: [Value]});
+
+  //# Traversable :: TypeClass
+  //.
+  //. `TypeClass` value for [Traversable][].
+  //.
+  //. ```javascript
+  //. > Traversable.test([])
+  //. true
+  //.
+  //. > Traversable.test({})
+  //. false
+  //. ```
+  var Traversable = $('Traversable', [Functor, Foldable], {traverse: [Value]});
+
+  //# Extend :: TypeClass
+  //.
+  //. `TypeClass` value for [Extend][].
+  //.
+  //. ```javascript
+  //. > Extend.test([])
+  //. true
+  //.
+  //. > Extend.test({})
+  //. false
+  //. ```
+  var Extend = $('Extend', [], {extend: [Value]});
+
+  //# Comonad :: TypeClass
+  //.
+  //. `TypeClass` value for [Comonad][].
+  //.
+  //. ```javascript
+  //. > Comonad.test(Identity(0))
+  //. true
+  //.
+  //. > Comonad.test([])
+  //. false
+  //. ```
+  var Comonad = $('Comonad', [Functor, Extend], {extract: [Value]});
+
+  //  Null$prototype$toString :: Null ~> () -> String
+  var Null$prototype$toString = function() {
+    return 'null';
+  };
+
+  //  Null$prototype$equals :: Null ~> Null -> Boolean
+  var Null$prototype$equals = function(other) {
+    return true;
+  };
+
+  //  Undefined$prototype$toString :: Undefined ~> () -> String
+  var Undefined$prototype$toString = function() {
+    return 'undefined';
+  };
+
+  //  Undefined$prototype$equals :: Undefined ~> Undefined -> Boolean
+  var Undefined$prototype$equals = function(other) {
+    return true;
+  };
+
+  //  Boolean$prototype$toString :: Boolean ~> () -> String
+  var Boolean$prototype$toString = function() {
+    return typeof this === 'object' ?
+      'new Boolean(' + toString(this.valueOf()) + ')' :
+      this.toString();
+  };
+
+  //  Boolean$prototype$equals :: Boolean ~> Boolean -> Boolean
+  var Boolean$prototype$equals = function(other) {
+    return typeof other === typeof this && other.valueOf() === this.valueOf();
+  };
+
+  //  Number$prototype$toString :: Number ~> () -> String
+  var Number$prototype$toString = function() {
+    return typeof this === 'object' ?
+      'new Number(' + toString(this.valueOf()) + ')' :
+      1 / this === -Infinity ? '-0' : this.toString(10);
+  };
+
+  //  Number$prototype$equals :: Number ~> Number -> Boolean
+  var Number$prototype$equals = function(other) {
+    return typeof other === 'object' ?
+      typeof this === 'object' &&
+        equals(other.valueOf(), this.valueOf()) :
+      isNaN(other) && isNaN(this) ||
+        other === this && 1 / other === 1 / this;
+  };
+
+  //  Date$prototype$toString :: Date ~> () -> String
+  var Date$prototype$toString = function() {
+    var x = isNaN(this.valueOf()) ? NaN : this.toISOString();
+    return 'new Date(' + toString(x) + ')';
+  };
+
+  //  Date$prototype$equals :: Date ~> Date -> Boolean
+  var Date$prototype$equals = function(other) {
+    return equals(other.valueOf(), this.valueOf());
+  };
+
+  //  RegExp$prototype$equals :: RegExp ~> RegExp -> Boolean
+  var RegExp$prototype$equals = function(other) {
+    return other.source === this.source &&
+           other.global === this.global &&
+           other.ignoreCase === this.ignoreCase &&
+           other.multiline === this.multiline &&
+           other.sticky === this.sticky &&
+           other.unicode === this.unicode;
+  };
+
+  //  String$empty :: () -> String
+  var String$empty = function() {
+    return '';
+  };
+
+  //  String$prototype$toString :: String ~> () -> String
+  var String$prototype$toString = function() {
+    return typeof this === 'object' ?
+      'new String(' + toString(this.valueOf()) + ')' :
+      '"' + this.replace(/\\/g, '\\\\')
+                .replace(/[\b]/g, '\\b')  // \b matches word boundary;
+                .replace(/\f/g, '\\f')    // [\b] matches backspace
+                .replace(/\n/g, '\\n')
+                .replace(/\r/g, '\\r')
+                .replace(/\t/g, '\\t')
+                .replace(/\v/g, '\\v')
+                .replace(/\0/g, '\\0')
+                .replace(/"/g, '\\"') + '"';
+  };
+
+  //  String$prototype$equals :: String ~> String -> Boolean
+  var String$prototype$equals = function(other) {
+    return typeof other === typeof this && other.valueOf() === this.valueOf();
+  };
+
+  //  String$prototype$concat :: String ~> String -> String
+  var String$prototype$concat = function(other) {
+    return this + other;
+  };
+
+  //  Array$empty :: () -> Array a
+  var Array$empty = function() {
+    return [];
+  };
+
+  //  Array$of :: a -> Array a
+  var Array$of = function(x) {
+    return [x];
+  };
+
+  //  Array$chainRec :: ((a -> c, b -> c, a) -> Array c, a) -> Array b
+  var Array$chainRec = function(f, x) {
+    var next = function(x) { return {value: x, done: false}; };
+    var done = function(x) { return {value: x, done: true}; };
+    var $todo = [x];
+    var $done = [];
+    while ($todo.length > 0) {
+      var xs = f(next, done, $todo.shift());
+      var $more = [];
+      for (var idx = 0; idx < xs.length; idx += 1) {
+        (xs[idx].done ? $done : $more).push(xs[idx].value);
+      }
+      Array.prototype.unshift.apply($todo, $more);
+    }
+    return $done;
+  };
+
+  //  Array$prototype$toString :: Array a ~> () -> String
+  var Array$prototype$toString = function() {
+    var reprs = this.map(toString);
+    var keys = Object.keys(this).sort();
+    for (var idx = 0; idx < keys.length; idx += 1) {
+      var k = keys[idx];
+      if (!/^\d+$/.test(k)) {
+        reprs.push(toString(k) + ': ' + toString(this[k]));
+      }
+    }
+    return '[' + reprs.join(', ') + ']';
+  };
+
+  //  Array$prototype$equals :: Array a ~> Array a -> Boolean
+  var Array$prototype$equals = function(other) {
+    if (other.length !== this.length) return false;
+    for (var idx = 0; idx < this.length; idx += 1) {
+      if (!equals(other[idx], this[idx])) return false;
+    }
+    return true;
+  };
+
+  //  Array$prototype$concat :: Array a ~> Array a -> Array a
+  var Array$prototype$concat = function(other) {
+    return this.concat(other);
+  };
+
+  //  Array$prototype$map :: Array a ~> (a -> b) -> Array b
+  var Array$prototype$map = function(f) {
+    return this.map(function(x) { return f(x); });
+  };
+
+  //  Array$prototype$ap :: Array a ~> Array (a -> b) -> Array b
+  var Array$prototype$ap = function(fs) {
+    var result = [];
+    for (var idx = 0; idx < fs.length; idx += 1) {
+      for (var idx2 = 0; idx2 < this.length; idx2 += 1) {
+        result.push(fs[idx](this[idx2]));
+      }
+    }
+    return result;
+  };
+
+  //  Array$prototype$chain :: Array a ~> (a -> Array b) -> Array b
+  var Array$prototype$chain = function(f) {
+    var result = [];
+    this.forEach(function(x) { Array.prototype.push.apply(result, f(x)); });
+    return result;
+  };
+
+  //  Array$prototype$reduce :: Array a ~> ((b, a) -> b, b) -> b
+  var Array$prototype$reduce = function(f, initial) {
+    return this.reduce(function(acc, x) { return f(acc, x); }, initial);
+  };
+
+  //  Array$prototype$traverse :: Applicative f => Array a ~> (a -> f b, c -> f c) -> f (Array b)
+  var Array$prototype$traverse = function(f, of) {
+    var applicative = of([]);
+    for (var idx = this.length - 1; idx >= 0; idx -= 1) {
+      applicative = ap(map(prepend, f(this[idx])), applicative);
+    }
+    return applicative;
+  };
+
+  //  Array$prototype$extend :: Array a ~> (Array a -> b) -> Array b
+  var Array$prototype$extend = function(f) {
+    return [f(this)];
+  };
+
+  //  Arguments$prototype$toString :: Arguments ~> String
+  var Arguments$prototype$toString = function() {
+    var args = Array.prototype.map.call(this, toString).join(', ');
+    return '(function () { return arguments; }(' + args + '))';
+  };
+
+  //  Arguments$prototype$equals :: Arguments ~> Arguments -> Boolean
+  var Arguments$prototype$equals = function(other) {
+    return Array$prototype$equals.call(this, other);
+  };
+
+  //  Error$prototype$toString :: Error ~> () -> String
+  var Error$prototype$toString = function() {
+    return 'new ' + this.name + '(' + toString(this.message) + ')';
+  };
+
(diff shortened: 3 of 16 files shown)


kedashoe:
> The last option seems best to me. We'll have all the definitions of Fantasy Land methods for built-in types in one place rather than scattered through a dozen or so files in Ramda. :)

So sanctuary (or whomever) can use these implementations to help with FL compliance for built in types? For example `map` could be used here: https://github.com/sanctuary-js/sanctuary/pull/142/files#diff-168726dbe96b3ce427e7fedce31bb0bcR579?

index.js
```diff
@@ -0,0 +1,306 @@
...
+    concat: 'concat',
+    empty: 'empty',
+    equals: 'equals',
+    extend: 'extend',
+    extract: 'extract',
+    map: 'map',
+    of: 'of',
+    reduce: 'reduce',
+    sequence: 'sequence'
+  };
```
Avaq: Why not depend on [`fantasy-land`](https://github.com/fantasyland/fantasy-land/blob/master/index.js) to get these constants? A change to them would probably be released as a major version, and you can always do like I did; [provide *both* the "easy" names, and the fantasy-land prefixed names](https://github.com/Avaq/Fluture/blob/master/fluture.js#L498-L521).

davidchambers: Three things need to happen before this is an option. I just opened fantasyland/fantasy-land#139 for discussion. :)

davidchambers:
> So sanctuary (or whomever) can use these implementations to help with FL compliance for built in types?

That's right! :)

> For example `map` could be used here: https://github.com/sanctuary-js/sanctuary/pull/142/files#diff-168726dbe96b3ce427e7fedce31bb0bcR579?

It's better than that, even. The Functor instances for `Array a`, `StrMap a`, and `a -> b` will be defined in this repository. `STC.map` will dispatch to one of these, or dispatch to a `map` method when applied to a value of another type. To define `S.map`, we'll simply use `def` to wrap `STC.map` (for currying and type checking). In other words, the logic for `map` will live here rather than in Sanctuary itself. The nice thing is that we can define, document, and test each Functor instance individually, whereas in Sanctuary we'd end up with a `switch` block or similar.

index.js
```diff
@@ -0,0 +1,471 @@
...
+  //. | [Monoid][]      |     ✔︎     |           |     ✔︎     |     ✔︎     |
+  //. | [Functor][]     |     ✔︎     |     ✔︎     |     ✔︎     |           |
+  //. | [Foldable][]    |     ✔︎     |           |           |           |
+  //. | [Traversable][] |     ✔︎     |           |           |           |
+  //. | [Apply][]       |     ✔︎     |     ✔︎     |           |           |
+  //. | [Applicative][] |           |           |           |           |
+  //. | [Chain][]       |     ✔︎     |           |           |           |
+  //. | [Monad][]       |           |           |           |           |
+  //. | [Extend][]      |           |           |           |           |
+  //. | [Comonad][]     |           |           |           |           |
```
davidchambers: | Type class      |   Array   | Function  |  Object   |  String   |
|:--------------- |:---------:|:---------:|:---------:|:---------:|
| [Setoid][]      |           |           |           |           |
| [Semigroup][]   |     ✔︎     |           |     ✔︎     |     ✔︎     |
| [Monoid][]      |     ✔︎     |           |     ✔︎     |     ✔︎     |
| [Functor][]     |     ✔︎     |     ✔︎     |     ✔︎     |           |
| [Foldable][]    |     ✔︎     |           |           |           |
| [Traversable][] |     ✔︎     |           |           |           |
| [Apply][]       |     ✔︎     |     ✔︎     |           |           |
| [Applicative][] |           |           |           |           |
| [Chain][]       |     ✔︎     |           |           |           |
| [Monad][]       |           |           |           |           |
| [Extend][]      |           |           |           |           |
| [Comonad][]     |           |           |           |           |

Could we check any more of these cells? I think I've implemented all the ones Ramda does, plus `Object#concat` (which is known as `merge` in Ramda).


[Applicative]:    https://github.com/fantasyland/fantasy-land#applicative
[Apply]:          https://github.com/fantasyland/fantasy-land#apply
[Chain]:          https://github.com/fantasyland/fantasy-land#chain
[Comonad]:        https://github.com/fantasyland/fantasy-land#comonad
[Extend]:         https://github.com/fantasyland/fantasy-land#extend
[Foldable]:       https://github.com/fantasyland/fantasy-land#foldable
[Functor]:        https://github.com/fantasyland/fantasy-land#functor
[Monad]:          https://github.com/fantasyland/fantasy-land#monad
[Monoid]:         https://github.com/fantasyland/fantasy-land#monoid
[Semigroup]:      https://github.com/fantasyland/fantasy-land#semigroup
[Setoid]:         https://github.com/fantasyland/fantasy-land#setoid
[Traversable]:    https://github.com/fantasyland/fantasy-land#traversable

svozza:
I'm only after getting a chance to look into this now. I didn't appreciate fully what was going on but now I've read the code I really like it. Generalising the dispatching seems like it will be really powerful. This is great.

index.js
```diff
@@ -0,0 +1,471 @@
...
+    return function(x) {
+      return apply(x)(f(x));
+    };
+  };
+
+  //. ### Object
+
+  //# Object$concat :: StrMap a ~> StrMap a
+  methods.Object$concat = function(strMap) {
+    var result = {};
```
svozza: This implementation is a bit hard to follow, does it really need to depend on `Array$map`? What's actually happening here?

davidchambers: We're merging two values of type `StrMap a`. Perhaps this implementation is clearer:

```javascript
var result = [];
for (var k in this) if (hasOwnProperty_.call(this, k)) result[k] = this[k];
for (k in strMap) if (hasOwnProperty_.call(strMap, k)) result[k] = strMap[k];
return result;
```

svozza: Yeah, I much prefer that implementation.

index.js
```diff
@@ -0,0 +1,471 @@
...
+      }
+    }
+    return result;
+  };
+
+  //# Array$chain :: Array a ~> (a -> Array b) -> Array b
+  methods.Array$chain = function(f) {
+    var result = [];
+    for (var idx = 0; idx < this.length; idx += 1) {
+      push_.apply(result, f(this[idx]));
```
svozza: Nice...

index.js
davidchambers: :zap:

svozza:
Any reason why there's no `Array$of`?

davidchambers:
> Any reason why there's no `Array$of`?

I had it initially, but then realized that `λ.of` wouldn't be very useful:

```javascript
> λ.of([], 42)
[42]

> λ.of(S.Nothing(), 42)
Just(42)

> λ.of(S.Right(0), 42)
Right(42)
```

Certain functions, such as `sequence`, take `of :: Applicative f => a -> f a` as an argument, which is more direct than taking a value with an `of` method (or a value with a `constructor` property with a static `of` method). To such functions one provides `R.of`<sup>†</sup>, `S.Maybe.of`, `S.Either.of`, or similar.

---

† _We could define `S.of` as the sanctified equivalent of this function._

svozza:
But do we not need `of` to define Array as an Applicative?

davidchambers:
> But do we not need `of` to define Array as an Applicative?

Great point, Stefano! The `eq(λ.Applicative.test([]), true)` assertion passed for me locally because my Node version (v4.4.3) defines [`Array.of`][1]. I've defined `methods.Array$of` to avoid this dependency. :)


[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/of

index.js
svozza: Can we tick the applicative box for arrays?

davidchambers: :zap:

index.js
rjmk: That's a shame -- any reason why sanctuary-js projects won't delegate to bundlers like browserify?

davidchambers: > any reason why sanctuary-js projects won't delegate to bundlers like browserify?

To what end, @rjmk? I'm certainly open to doing so, but it's not clear to me exactly what we stand to gain. I haven't done much front-end development in the past two and a half years, so I'm a bit out of touch. :stuck_out_tongue_winking_eye:

rjmk: Well, it gives control of how to bundle the code to the consumer. It also allows us to be much more flexible in which modules we use (as here). It makes the code cleaner.

[It is trivial](https://wzrd.in) to generate standalone bundles from npm modules, and so you don't even need to be using a bundler to use an npm package on the front-end.

If we wanted to have a bower version available, we could bundle sanctuary-js projects into a `dist/` folder.

davidchambers: This all sounds reasonable to me. What about AMD? I just want to be clear how all the pieces fit together (and some of the pieces in question are foreign to me).

rjmk: I've never used AMD, but the bundles produced by wzrd, for example, are AMD compatible

davidchambers: Okay. I've converted the file to CommonJS format and added `fantasy-land` as a dependency.

davidchambers:
I've spent significant time over the past two days reviving this pull request. There's been a lot of activity in Fantasy Land since I opened this pull request, so there was quite a lot of catching up to do.

This pull request still isn't quite ready to merge, but it's ready for review.

For each Fantasy Land method I'd like to provide implementations for as many built-in types as possible. Let me know if you spot an omission!

safareli:
There is no `Z.chainRec` maybe you have forgotten

safareli:
Also there is no implementation of ChainRec for Array, if something is Chain it could also be ChainRec. [Here I have implemented](https://gist.github.com/safareli/ea45b45bed99f32bae7002e19dcf13c2) it which you could use for reference

davidchambers:
I just replaced `sequence` with `traverse`.

> There is no `Z.chainRec` maybe you have forgotten

It's on my to-do list, but thanks for the reminder. I'll add it later today. :)

> Also there is no implementation of ChainRec for Array, if something is Chain it could also be ChainRec. [Here I have implemented](https://gist.github.com/safareli/ea45b45bed99f32bae7002e19dcf13c2) it which you could use for reference

Excellent! I'll have a look at your implementation and define `Array#fantasy-land/chainRec`. :)

index.js
```diff
@@ -0,0 +1,680 @@
...
+    },
+
+    //  Array#traverse :: Applicative f => Array a ~> ((a -> f b), (c -> f c)) -> f (Array b)
+    'Array#fantasy-land/traverse': function(f, of) {
+      //  prepend :: a -> Array a -> Array a
+      var prepend = function(x) {
+        return function(xs) {
+          return [x].concat(xs);
+        };
+      };
```
Avaq: Maybe move `prepend` out and up to the other utilities? It does not seem to need any of its enclosed variables, it seems quite general even if it's only used once, and moving it should stop the function from being allocated every time `traverse` is called.

test/index.js
```diff
@@ -0,0 +1,359 @@
...
+    eq(Z.Monoid.test({}), true);
+    eq(Z.Monoid.test(''), true);
+  });
+
+  it('Functor.test', function() {
+    eq(Z.Functor.test(null), false);
+    eq(Z.Functor.test([]), true);
+    eq(Z.Functor.test({}), true);
+    eq(Z.Functor.test(''), false);
+    eq(Z.Functor.test(inc), true);
```
safareli: should it also includ `map|fantasy-land/map`?

index.js
```diff
@@ -0,0 +1,680 @@
...
+    'Object#fantasy-land/empty': function() { return {}; },
+
+    //  Object#map :: StrMap a ~> (a -> b) -> StrMap b
+    'Object#fantasy-land/map': function(f) {
+      var result = {};
+      for (var k in this) if (has(k, this)) result[k] = f(this[k]);
+      return result;
+    },
+
+    //  String#toString :: String -> String
```
safareli: I think it should be `~>` instead of  `->` ?

index.js
davidchambers: Good idea. I've just made your suggested change. :)

Thanks for taking the time to review this pull request. I'm pleased you like the approach to dispatching.

index.js
davidchambers: Great catch! Fixed. :)

davidchambers:
> would it be better to depend on fantasy-land and instead of string `fantasy-land/method` use `fl.method`?

I've vacillated on this one. I depended on fantasy-land until a few days ago. My reasons for favouring string literals are as follows:

  - __Bower support.__ Sanctuary currently works in the browser and can be installed via Bower. As of fantasyland/fantasy-land#153, fantasy-land works in the browser, but `bower install fantasy-land` is not yet possible (see fantasyland/fantasy-land#139). If and when this becomes possible we should revisit this matter.

  - __Object literals.__ I enjoy being able to write…

    ```javascript
    return {
      'fantasy-land/equals': ...,
      'fantasy-land/map': ...,
      'fantasy-land/ap': ...,
      ...
    };
    ```

    rather than…

    ```javascript
    var o = {};
    o[fl.equals] = ...;
    o[fl.map] = ...;
    o[fl.ap] = ...;
    return o;
    ```

    I realize ES2015 provides support for computed keys in object literals, but we're not yet using a JavaScript transpiler. If and when we adopt one we should revisit this matter.

  - __Dependencies.__ All else being equal, the fewer dependencies a project has the better.

Reasons to change my mind:

  - __Maintenance.__ The main reason for favouring `fl.map` over `'fantasy-land/map'` is that the value of `fl.map` may one day change. If one has hard-coded the value in multiple places, upgrading to the new version of fantasy-land would require more effort. I think it's *highly unlikely*, though, that these values will ever change, so I don't find this reason compelling.

  - __Explicitness.__ The nice thing about depending on fantasy-land is that __package.json__ would state which version of the spec we support. We can of course state this in the documentation, but an entry in the `"dependencies"` map could not become out of date.

I find the arguments on both sides quite weak. Two of the three reasons for favouring string literals may disappear in the future, and the third reason is offset by a compensating benefit (making it clear which version of the spec we support).

The most significant point does not belong in either list:

  - __Correctness.__ One could mistype `'fantasy-land/map'`, but one could also mistype `fl.map` (though not so easily). A misspelt property name could not be detected statically without the help of an ESLint plugin. In order to have the linter catch typos, I had this at one point:

    ```javascript
    const AP = FantasyLand.ap;
    const CHAIN = FantasyLand.chain;
    ...
    ```

    This felt like a lot of boilerplate. I decided instead to have Vim highlight the fantasy-land strings as constants:

    ```
    highlight link FantasyLand Constant
    syntax match FantasyLand +\v(['"])\zsfantasy-land/(ap|bimap|chain|chainRec|concat|empty|equals|extend|extract|map|of|promap|reduce|traverse)(\1)@=+ containedin=jsString,jsObjectStringKey
    ```

    This has worked well for me so far. After fantasyland/fantasy-land#135 I replaced `sequence` with `traverse` in __~/.vim/ftplugin/javascript.vim__. When I switched back to working on this project I noticed that occurrences of `'fantasy-land/sequence'` were no longer highlighted like all the other fantasy-land strings. :)

    We should revisit this matter if and when someone releases an ESLint plugin for validating property access of the object acquired via `require('fantasy-land')`.

> I could not see `Z.of` is it intentional?

Upon rereading https://github.com/sanctuary-js/sanctuary-type-classes/pull/2#issuecomment-218796444 I'm no longer convinced of my prior position. Requiring the caller to provide `of` is fine when defining a function specific to one type: one can use `R.of` for Array, `S.Just` for Maybe, etc. If, though, one wishes to define a function for all applicatives one would like to get `of` from a type dictionary. One might end up writing

```javascript
typeof applicative['fantasy-land/of'] === 'function' ?
  applicative['fantasy-land/of'] :
  applicative.constructor['fantasy-land/of']
```

in order to "find" the `of` function. Perhaps `Z.of` could take care of this. Is this what you have in mind, @safareli?

```javascript
//# of :: Applicative f => f a -> b -> f b
var of = function(applicative, x) {
  return applicative.constructor != null &&
         typeof applicative.constructor['fantasy-land/of'] === 'function' ?
    applicative.constructor['fantasy-land/of'](x) :
    dispatch('fantasy-land/of', applicative, [x]);
};
Z.of = of;
```

test/index.js
davidchambers: The problem is that having defined `Object#fantasy-land/map`, all plain JavaScript objects satisfy the predicate. I did add these assertions, though:

```javascript
eq(Z.Functor.test({'@@type': 'my-package/Foo', map: method}), false);
eq(Z.Functor.test({'@@type': 'my-package/Foo', 'fantasy-land/map': null}), false);
eq(Z.Functor.test({'@@type': 'my-package/Foo', 'fantasy-land/map': method}), true);
```

All the `<Type>.test` assertions were useful during development, but I believe they've outlived their usefulness. `eq(Z.Chain.test([]), true)` only asserts that `methods` contains an entry for `Array#fantasy-land/chain`. We could assert this indirectly by testing `Z.chain`.

What I'd like to do is write more tests for `Z.TypeClass` so we can have enough confidence in

```javascript
var Chain = Z.Chain =
TypeClass('sanctuary-type-classes/Chain',
          [Apply],
          hasMethod('fantasy-land/chain'));
```

that we're not obliged to test `Z.Chain.test` and all the other predicates individually.

safareli:
Yes as we have `empty` i think we should also have `of` for completeness.

We could state fantasy-land in Peer and Dev Dependencies and have umd build for browsers but it could be done in other PR

index.js
```diff
@@ -0,0 +1,666 @@
...
+    //  Object#empty :: StrMap a ~> StrMap a
+    'Object#fantasy-land/empty': function() { return {}; },
+
+    //  Object#map :: StrMap a ~> (a -> b) -> StrMap b
+    'Object#fantasy-land/map': function(f) {
+      var result = {};
+      for (var k in this) if (has(k, this)) result[k] = f(this[k]);
+      return result;
+    },
+
```
kedashoe: Is there a reason not to have more `Object` methods? Having `empty`, `concat` and `map`.. couldn't we have the others as well? For example, ramda not supporting `reduce` for objects always gets me. I have an application with a lot of `StrMap Number`s, and [I cannot](https://github.com/ramda/ramda/blob/master/src/sum.js#L20) use `sum` in general functions that I may at some point pass an array or an object to.

safareli:
Small nit, I think it would be better to create separate commits and then squish in one if you prefer before merging as it's hard to follow changes.

davidchambers:
> I think it would be better to create separate commits and then squish in one if you prefer before merging as it's hard to follow changes.

Good idea. I'll do this from now on.

index.js
davidchambers: I like the idea of `Object#fantasy-land/reduce`. Does the binary operation need to be both commutative and associative in order for the result to be deterministic?

rjmk: `g = a => b => c => f(a, f(b, c))`, where `f` is the reducing function `:: a -> b -> b`, has to be commutative in its first two arguments.

davidchambers:
One thing that bothered me was the similarity between the code which determined whether a value satisfies the requirements of Monoid, say, and the code for dispatching to the corresponding method (`empty`, in this case). In one case we simply want to know whether such a method exists; in the other we actually want to invoke it. It would be nice to share the lookup logic.

Let's start by specifying which methods are required and where they may be defined:

```javascript
//  data Location = Constructor | Value

//  Constructor :: Location
var Constructor = 'constructor';

//  Value :: Location
var Value = 'value';
```

```javascript
var Monoid = Z.Monoid =
FantasyLandTypeClass('sanctuary-type-classes/Monoid',
                     [Semigroup],
                     {empty: [Value, Constructor]});
```

The crucial part is `{empty: [Value, Constructor]}`. This says that in order to satisfy Monoid a value must—in addition to satisfying Semigroup—provide a `fantasy-land/empty` method on itself (`Value`) or on its constructor object (`Constructor`).

From this we're able to derive two things: a predicate to provide to `Z.TypeClass`, and a function for accessing a value's `fantasy-land/empty` method. The order of the "locations" determines the order in which the lookups are performed.

`FantasyLandTypeClass` returns a `TypeClass` value with an extra field, `getBoundMethod`, which takes a method name and a value and returns the method with `this` bound to the value.

With this machinery in place the implementation of `empty` is trivial:

```javascript
//  empty :: Monoid m => m -> m
var empty = function(x) {
  return Monoid.getBoundMethod('empty', x)();
};
```

safareli:
Nice! looks good

index.js
```diff
@@ -0,0 +1,661 @@
...
+    //  Array#map :: Array a ~> (a -> b) -> Array b
+    'Array#fantasy-land/map': function(f) {
+      return this.map(function(x) { return f(x); });
+    },
+
+    //  Array#ap :: Array a ~> Array (a -> b) -> Array b
+    'Array#fantasy-land/ap': function(fs) {
+      var xs = this;
+      return chain(function(f) { return map(f, xs); }, fs);
+    },
```
davidchambers: I just updated this to remove the nested `for` loops of the earlier implementation. :)

index.js
```diff
@@ -432,7 +465,10 @@
     'String#fantasy-land/concat': function(s) { return this + s; },
 
     //  String#empty :: String ~> String
-    'String#fantasy-land/empty': function() { return ''; }
+    'String#fantasy-land/empty': function() { return ''; },
+
+    //  Undefined#equals :: Undefined ~> Undefined -> Boolean
+    'Undefined#fantasy-land/equals': function(other) { return true; }
```
safareli: Why do we ignore `other`? what if it's not `undefined`? (same with `Null`)
looks like tests are correct but I don't understand why :D

davidchambers: `Z.equals` contains the guard `type(x) === type(y)`. If `x` and `y` are not members of the same type we don't dispatch to `fantasy-land/equals`. This means that each equality function is only responsible for comparing two members of its type. For example:

```haskell
Undefined#fantasy-land/equals :: Undefined ~> Undefined -> Boolean
```

There is only one member of the Undefined type, so checking the value is unnecessary. We could use `other === undefined` instead, but it would then be easy to assume that the method's domain is all values rather than all values of the Undefined type.

safareli: makes sense!

index.js
```diff
@@ -0,0 +1,729 @@
...
+  var chain = function(f, chain) {
+    return Chain.getBoundMethod('chain', chain)(f);
+  };
+  Z.chain = chain;
+
+  //# chainRec :: ChainRec m => (m x, (a -> c, b -> c, a) -> m c, a) -> m b
+  var chainRec = function(chainRec, f, x) {
+    return ChainRec.getBoundMethod('chainRec', chainRec)(f, x);
+  };
+  Z.chainRec = chainRec;
```
davidchambers: @safareli and others, with partial application in mind what is the best argument order for `Z.chainRec`? It's clear to me that `f` should come before `x`, leaving three possibilities:

  - `(chainRec, f, x)`
  - `(f, chainRec, x)`
  - `(f, x, chainRec)`

If `f` uses Array as its ChainRec, `chainRec` must be an array. Thus the third possibility isn't useful: one would never wish to provide `f` without also providing a matching `chainRec`.

There *is* value, though, in providing `chainRec` without `f`. One could write:

```javascript
const chainRecArray = Z.chainRec([]);
```

`(chainRec, f, x)` appears to be the best order. Do you agree? Have I overlooked something important?

safareli: Agree on `(chainRec, f, x)`

index.js
```diff
@@ -0,0 +1,734 @@
...
+  //. the exported `map` function dispatches when appropriate.
+  //.
+  //. | Type class      |   Array   | Function  |  Object   |  String   |
+  //. |:--------------- |:---------:|:---------:|:---------:|:---------:|
+  //. | [Setoid][]      |     ✔︎     |     ✔︎     |     ✔︎     |     ✔︎     |
+  //. | [Semigroup][]   |     ✔︎     |           |     ✔︎     |     ✔︎     |
+  //. | [Monoid][]      |     ✔︎     |           |     ✔︎     |     ✔︎     |
+  //. | [Functor][]     |     ✔︎     |     ✔︎     |     ✔︎     |           |
+  //. | [Bifunctor][]   |           |           |           |           |
+  //. | [Profunctor][]  |           |           |           |           |
```
safareli: I think this table should be updated as we have bimap for Function

index.js
```diff
@@ -0,0 +1,734 @@
...
+      var done = function(x) { return {value: x, done: true}; };
+      var $todo = [x];
+      var $done = [];
+      while ($todo.length > 0) {
+        var $buckets = [[], []];
+        f(next, done, $todo.shift()).forEach(function(x) {
+          $buckets[x.done ? 1 : 0].push(x.value);
+        });
+        unshift_.apply($todo, $buckets[0]);
+        push_.apply($done, $buckets[1]);
```
safareli: what's the benefit of having this `bucket`? I think we could directly unshift/push in forEach like this:
```js
f(next, done, $todo.shift()).forEach(function(x) {
  if (x.done) {
    $done.push(x.value);
  } else {
    $todo.unshift(x.value);
  }
});
```
or even inline `if/else` like this:
```js
f(next, done, $todo.shift()).forEach(function(x) {
  x.done ? $done.push(x.value) : $todo.unshift(x.value);
});
```

safareli: Also I think it would be better to declare the function before `while` loop, or use `for`.

index.js
davidchambers: Good catch (though I think you meant `promap`). I've updated the table. :)

index.js
davidchambers: I tried your suggestion initially, but `Array.prototype.unshift.apply(xs, [1, 2, 3])` and `xs.unshift(1); xs.unshift(2); xs.unshift(3)` are not equivalent (even ignoring mutation) because the elements are added backwards with the second approach.

I like the idea of using a `for` loop here. I'll push a commit shortly.

safareli: ah, got it.
but we can do loop from back like this:
```
Array.chainRec = function(f, i) {
  var next = function(x) { return {value: x, done: false}; };
  var done = function(x) { return {value: x, done: true}; };
  var $todo = [i];
  var $done = [];
  var i, xs, x;
  while ($todo.length > 0) {
    xs = f(next, done, $todo.shift());
    for (i = xs.length -1; i >= 0; i--) {
      x = xs[i]
      if (x.done) {
        $done.push(x.value);
      } else {
        $todo.unshift(x.value);
      }
    }
  }
  return $done;
}

function squash(next, done, a) {
  return Array.isArray(a) ? a.map(next) : [done(a)];
}

console.log(Array.chainRec(squash, [1, [[2, 3], 4], 5])); // [ 1, 2, 3, 4, 5 ]
```

safareli: This way we could avoid:
- creating temporary `$more` array 
- allocation of new array on `concat`

davidchambers: Looping backwards seems to introduce a new problem: we get `['aaa?', 'aaa!', 'aab?', …]` rather than `['aaa!', 'aaa?', 'aab!', …]`.

safareli: this will do the trick 😬
```diff
Array.chainRec = function _chainRec(f, i) {
  var next = function(x) { return {value: x, done: false}; };
  var done = function(x) { return {value: x, done: true}; };
  var $todo = [i];
  var $done = [];
  var xs, x, idx;
  while ($todo.length > 0) {
-    xs = f(next, done, $todo.shift());
+    xs = f(next, done, $todo.pop());
    for (idx = xs.length - 1; idx >= 0; idx -= 1) {
      x = xs[idx];
      if (x.done) {
-        $done.push(x.value);      
+        $done.unshift(x.value);
      } else {
        $todo.unshift(x.value);
      }
    }
  }
  return $done;
}

```

safareli: As we are doing `unshift` in all cases it could be:
```js
function _chainRec(f, i) {
  var next = function(x) { return {value: x, done: false}; };
  var done = function(x) { return {value: x, done: true}; };
  var $todo = [i];
  var $done = [];
  var xs, idx;
  while ($todo.length > 0) {
    xs = f(next, done, $todo.pop());
    for (idx = xs.length - 1; idx >= 0; idx -= 1) {
      (xs[idx].done ? $done : $todo).unshift(xs[idx].value);
    }
  }
  return $done;
}
```

safareli: never mind it's not working with `squash`

davidchambers: This version fails a different test: the output is `[2, 3, 4, 1, 5]` rather than `[1, 2, 3, 4, 5]` for the `squash` test. You're welcome to clone the repository, run `make setup`, make changes, then run `make test` to verify that they pass the handful of tests we've added so far.

safareli: I only had test for `ab!?` in ramda fork and i thought it worked. but then added test for `squish` and it failed as you described.

After I thought about the implementation and realized that we would need to buffer items even if we try to do it in other direction.

I would just suggest this, and the implementation looks good to me:
```diff
-$todo = $more.concat($todo);
+unshift_.apply($todo, $more);
```

safareli:
`Type-class hierarchy` is so nice!!
you should add it to fantasy land instead of that outdated image.

safareli:
There was no example and link to FL of `chainRec` in readme.
We could use the `ab!?` and `squash` example provided by @scott-christopher:
```js
chainRec(Array, (next, done, a) => (
  a.length == 2 ? [a + '!', a + '?'].map(done)
                : [a + 'a', a + 'b'].map(next)
), ''); // [ 'aa!', 'aa?', 'ab!', 'ab?', 'ba!', 'ba?', 'bb!', 'bb?' ],

chainRec(Array, (next, done, a) => (
  Array.isArray(a) ? a.map(next) : [done(a)]
), [1, [[2, 3], 4], 5]); // [1, 2, 3, 4, 5]
```

Also not quite sure but I think we could make function and object Bifunctor.
- in case of function first will map it's input last it's output.
- in case of object firs will map it's key and last it's value.
(or in other way). 
But this is not the major thing and it could be done in other PR.

davidchambers:
> `Type-class hierarchy` is so nice!!
> You should add it to fantasy land instead of that outdated image.

Thanks! I agree: fantasyland/fantasy-land#160. It's a pity that it doesn't render nicely on mobile devices, bad that doesn't dissuade me from including it here. :)

> There was no example and link to FL of `chainRec` in readme.

Added. Thanks.

> Also not quite sure but I think we could make function and object Bifunctor.

I don't think Function is a candidate.

```haskell
bimap :: Bifunctor f => f a c ~> (a -> b) -> (c -> d) -> f b d
```

Replace `Bifunctor f =>` with `Function`:

```haskell
bimap :: Function a c ~> (a -> b) -> (c -> d) -> Function b d
```

Rewrite:

```haskell
bimap :: (a -> c) ~> (a -> b) -> (c -> d) -> (b -> d)
```

Given `a -> b`, `a -> c`, and `c -> d` there's no way to produce `b -> d` since none of the functions at our disposal takes a `b`.

Function can satisfy Profunctor, though. :)

I don't think Object is a candidate either. When we talk about an "object" we're either referring to a record of a specific type or a string map. [`StrMap :: Type -> Type`][1] has only one type parameter, but `bimap` requires a type with (at least) two type parameters. @scott-christopher suggested implementing `bimap` for the native Map type, but I don't want to encourage the use of native maps and sets because they're mutable and have identity-based equality. Eventually I'd like the Sanctuary ecosystem to contain immutable, FL-compatible Map and Set types.


[1]: https://github.com/sanctuary-js/sanctuary-def#strmap

README.md
```diff
@@ -252,7 +252,27 @@ See [`chain`][chain].
 
 <h4 name="chainRec"><code><a href="https://github.com/sanctuary-js/sanctuary-type-classes/blob/v0.0.0/index.js#L705">chainRec :: ChainRec m => (m x, (a -> c, b -> c, a) -> m c, a) -> m b</a></code></h4>
 
-<h4 name="reduce"><code><a href="https://github.com/sanctuary-js/sanctuary-type-classes/blob/v0.0.0/index.js#L710">reduce :: Foldable f => ((a, b) -> a) -> a -> f b -> a</a></code></h4>
+See [`chainRec`][chainRec].
+
+```javascript
+> chainRec(
+.   [],
```
safareli: I think it's much correct to pass Type (`Array` in this case) instead of creating throwaway object here.
Also it would be sort of confusing to newcomers as it's not clear why are we passing empty array.

index.js
```diff
@@ -0,0 +1,865 @@
...
+        //  Array#reduce :: Array a ~> ((b, a) -> b, b) -> b
+        'fantasy-land/reduce': function(f, initial) {
+          return this.reduce(function(acc, x) { return f(acc, x); }, initial);
+        },
+
+        //  Array#traverse :: Applicative f => Array a ~> ((a -> f b), (c -> f c)) -> f (Array b)
+        'fantasy-land/traverse': function(f, of) {
+          var applicative = of([]);
+          for (var idx = this.length - 1; idx >= 0; idx -= 1) {
+            applicative = ap(map(prepend, f(this[idx])), applicative);
```
svozza: This will be very slow:

https://github.com/ramda/ramda/issues/1709

davidchambers: Great catch! I'll investigate using a linked list tomorrow. :)

davidchambers: The performance actually seems reasonable. I did revert to an imperative implementation of `Array#fantasy-land/ap` which made a noticeable difference. Producing an array containing 10^6 arrays now takes less than one second on my computer. Using linked lists as suggested by @scott-christopher results in the same task taking slightly more than a second for me.

Here's the diff, for reference:

```diff
--- a/index.js
+++ b/index.js
@@ -132,6 +132,25 @@

   'use strict';

+  //  Nil :: List a
+  var Nil = {isNil: true, isCons: false};
+
+  //  Cons :: a -> List a -> List a
+  var Cons = function(head) {
+    return function(tail) {
+      return {isNil: false, isCons: true, head: head, tail: tail};
+    };
+  };
+
+  //  listToArray :: List a -> Array a
+  var listToArray = function(_list) {
+    var result = [];
+    for (var list = _list; list.isCons; list = list.tail) {
+      result.push(list.head);
+    }
+    return result;
+  };
+
   //  has :: (String, StrMap a) -> Boolean
   var has = function(k, strMap) {
     return Object.prototype.hasOwnProperty.call(strMap, k);
@@ -432,11 +451,11 @@

         //  Array#traverse :: Applicative f => Array a ~> ((a -> f b), (c -> f c)) -> f (Array b)
         'fantasy-land/traverse': function(f, of) {
-          var applicative = of([]);
+          var applicative = of(Nil);
           for (var idx = this.length - 1; idx >= 0; idx -= 1) {
-            applicative = ap(map(prepend, f(this[idx])), applicative);
+            applicative = ap(map(Cons, f(this[idx])), applicative);
           }
-          return applicative;
+          return map(listToArray, applicative);
         }
       }
     },
```

This expression produces an array containing 10^6 arrays:

```javascript
Z.traverse(R.of, R.identity, R.repeat(R.range(0, 10), 6))
```

I imagine there's still room for improvement. Feel free to experiment, @svozza. :)

svozza: Nice! I'll have a play with this on the weekend.

README.md
safareli: @davidchambers what you think on this?

davidchambers: This is a very good question. Let's discuss it in #3.

index.js
```diff
@@ -0,0 +1,979 @@
...
+  //
+  //  8.  If the method invocation in step 6 succeeded, its result is returned.
+  //      Otherwise, the exception is allowed to propagate.
+  //
+  //  9.  The ‘finally’ block in ‘toString’ pops the array pushed onto
+  //      $seenStack in step 1.
+  //
+  //  10. If the call to ‘show’ in step 2 succeeded, its result is returned.
+  //      Otherwise, the exception is allowed to propagate.
+  var $seenStack = [];
```
davidchambers: I'd love someone (@Avaq?) to verify that this approach is sound. It's necessarily different from the approach used for [`R.toString`][1]. In Ramda's case there's a single recursive `toString` function with one case for each type, whereas in our case each type has its own `toString` method. Since we can't expect a `toString` method to accept an array of "seen" values as an argument, we instead rely on "global" mutable state. This should be okay given JavaScript's single-threaded execution model.

Check out the uses of the wonderfully complex and surprising [`try...finally`][2] statement. I don't think I had ever used it before today!


[1]: http://ramdajs.com/0.22.1/docs/#toString
[2]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch

Avaq: :eyes:

Avaq: The idea of the `$seenStack` is to allow `toString` to receive nested calls, right? Otherwise the purpose of `$seenStack` is unclear to me. However, if we allow `toString` to be called like this, wouldn't there be the potential for it to encounter a value which has already been encountered in another `$seen`?

---

*Some time later:* I cloned the repo to see for myself, and to my surprise, deeply nested circular references are neatly detected and reported, but I realized that's because all nested structures are calling directly to `show`. However, there were rare cases where I found `$seenStack` to be larger than 1. After some more digging I realized these cases were caused by the custom type `test/Identity`, which implements `toString` which itself calls to `Z.toString`.

I then, unfortunately, broke it by:

```js
var o = {};
var id = Identity(o);
o.id = id;
Z.toString(id);
```

...which causes a stack overflow. :(

I will now ponder whether this is a problem, and what could be a solution.

davidchambers: Thank you for performing such a thorough investigation! I had a feeling you were the right person for the job. :)

> I will now ponder whether this is a problem, and what could be a solution.

I believe it's a problem, but I don't see a way to fix it without an API change. An elegant solution is to provide an `Any -> String` function as an argument to the method. At that point the method would be sufficiently different from a conventional `toString` method that it should probably have a different name; perhaps `show`. Let's consider `Identity#show`:

```javascript
//  Identity#show :: Identity a ~> (b -> String) -> String
Identity.prototype.show = function(show) {
  return 'Identity(' + show(this.value) + ')';
};
```

_Some time later:_ I think there *is* a way to handle the case you discovered without changing the API. Instead of looking for the value in the *last* array in `$seenStack`, we look to see whether it is in *any* array in `$seenStack`. This fixes your case without breaking any of the existing test cases, but I haven't yet convinced myself that it's sound. Could you pull the latest commit and try to break it again? :)

index.js
```diff
@@ -0,0 +1,954 @@
...
+
+    return function(x) {
+      if ($seen.indexOf(x) >= 0) return '<Circular>';
+
+      var xType = type(x);
+      var toStringMethod =
+        has(xType, implementations) &&
+        has('toString', implementations[xType].prototype) ?
+          implementations[xType].prototype.toString :
+          x.toString;
```
safareli: If user overrides toString on some type for which there is implementation `implementations ` then it would not be called.

I think it it's not intentional, we could store default implementations in `defaultImplementations` for example and here check if `x.toString` is same as one from `defaultImplementations` (if exists) then call `implementations` as default one is not overridden but if it's overridden then call `x.toString`. 

(we don't have this issue in `T.getBoundMethod` as it's firs looking method on instance)

davidchambers: Good catch! If `x` has a `@@type` property we'll use `x.toString`, but if not we'll use our `Object#toString` method even if `x` has a custom `toString` method. We shouldn't expect authors of algebraic data types to specify `@@type`, as this is (currently) just a Sanctuary convention. I hope one day Ramda and other libraries will depend on sanctuary-type-classes so the community has a canonical repository of implementations of Fantasy Land methods for built-in types. Requiring `@@type` would make this :rainbow: future less likely.

I'm not sure exactly what you're suggesting, though. If `x` is `{}`, say, we want to use our implementation rather than `x.toString`. We could use `x.toString` if it is not identical to `Object.prototype.toString`, but then we'd run into the problems noted in sanctuary-js/sanctuary#100. For example:

```javascript
> ({}).toString === Object.prototype.toString
true
> vm.runInNewContext('({})').toString === Object.prototype.toString
false
```

This leaves us trying to infer whether the function is `Object.prototype.toString`. This is how [Ramda][1] does it:

> ```javascript
> if (typeof x.toString === 'function') {
>   var repr = x.toString();
>   if (repr !== '[object Object]') {
>     return repr;
>   }
> }
> return '{' + mapPairs(x, keys(x)).join(', ') + '}';
> ```

The `'[object Object]'` check works well in practice, but I would love to discover a less hacky solution.


[1]: https://github.com/ramda/ramda/blob/v0.22.1/src/internal/_toString.js#L38-L44

test/index.js
```diff
@@ -295,15 +296,10 @@ suite('functions', function() {
...
-    eq(Z.empty(''), '');
-    eq(Z.empty('abc'), '');
-    eq(Z.empty([]), []);
-    eq(Z.empty([1, 2, 3]), []);
-    eq(Z.empty({}), {});
-    eq(Z.empty({x: 1, y: 2, z: 3}), {});
-    eq(Z.empty(Identity('abc')), Identity(''));
+    eq(Z.empty(String), '');
+    eq(Z.empty(Array), []);
+    eq(Z.empty(Object), {});
```
safareli: I think we can have `empty` as instance method and in that case it's a good test case:
```js
eq(Z.empty(Identity('abc')), Identity(''));
```

index.js
```diff
@@ -795,15 +813,13 @@
...
   //. .   [1, [[2, 3], 4], 5]
   //. . )
   //. [1, 2, 3, 4, 5]
   //. ```
-  var chainRec = function(chainRec, f, x) {
-    return ChainRec.getBoundMethod('chainRec', chainRec)(f, x);
-  };
+  var chainRec = ChainRec.staticMethods.chainRec;
 
   //# filter :: (Monad m, Monoid m a) => (a -> Boolean, m a) -> m a
```
safareli: I think we can make constraints on `m` weeker  ([src](https://wiki.haskell.org/Foldable_and_Traversable#Some_trickier_functions:_concatMap_and_filter)
```hs
filter :: (Applicative f, Foldable f, Monoid (f a)) => (a -> Bool) -> f a -> f a
filter p = foldMap (\a -> if p a then pure a else mempty)
```

index.js
safareli: What I thought about, goed down to reference check which would have the issue you have pointed out with vm.runInNewContext.

Ramda "hack" looks good best, but it also has issue:
```js
let Foo = function(val) {
  return {
    val: val,
    toString: function() {
      return (this.val == 0 )
        ? "Foo("+this.val+")"
        : Object.prototype.toString.call(this);
    },
  };
};

toString(Foo(0)) // "Foo(0)"
toString(Foo(1)) // "{\"toString\": function toString() {\n      return this.val == 0 ? \"Foo(\" + this.val + \")\" : Object.prototype.toString.call(this);\n    }, \"val\": 1}"
```

so question is:
- which one is more likely to happen?
- what are advantages/disadvantages of one or the other

test/index.js
davidchambers: I've updated `Z.empty` to take a type representative, so `Identity('abc')` is no longer a valid value. This is in line with fantasyland/fantasy-land#180, which will preclude the Identity type from satisfying the requirements of Monoid.

Initially I thought we should support both `Identity` and `Identity('abc')` by requiring an argument of type `Monoid m => { fantasy-land/empty :: () -> m }`, but we'd also need to support `Array` and `[1, 2, 3]`. The description of `Z.of` would be quite complex, and the type signature would not tell the full story.

`Monoid m => TypeRep m` is accurate and compatible with fantasyland/fantasy-land#176.

README.md
```diff
@@ -0,0 +1,348 @@
...
+.   (next, done, s) =>
+.     s.length === 2 ? [s + '!', s + '?'].map(done)
+.                    : [s + 'o', s + 'n'].map(next),
+.   ''
+. )
+['oo!', 'oo?', 'on!', 'on?', 'no!', 'no?', 'nn!', 'nn?']
+
+> chainRec(
+.   [],
```
safareli: Here Should be `Array` I think

davidchambers: Thanks. __README.md__ is generated by `make release-*`, so I'll actually delete the file before merging this pull request. It's useful to be able to view the rendered readme when reviewing the pull request, though, so I'll keep the file for the time being.

index.js
davidchambers: Interesting idea! Essentially we'd require `reduce` (from Foldable) rather than `chain` (from Chain). This may be a weaker requirement, but it's certainly not *strictly* weaker. Why do you believe Foldable to be easier to satisfy than Chain?

I made the change to confirm that we can do it. I pushed the commit.

davidchambers:
I opened this pull request five months ago. It was far from finished at the time; the description reflected this. I now realize I should have kept the description up to date so people could learn what this pull request is about from the title and description. I've finally updated the description. I'll do a better job of this in the future. :)

index.js
safareli: Actually you are write! We can't `reduce`: `Task`, `IO` or `Function`. But there are structures which can't be a Monad but are Foldable: Comonads (i.e. CoFree) and ZipList and maybe others. What if we have two filter functions like this maybe: `filter` and `filterM`

README.md
safareli: Ahh makes sense!

davidchambers:
This pull request is almost ready to merge, I believe. In preparation for merging I'll squash the commits into just two commits, the second of which will only update __README.md__. I'll remove this commit before merging, but it's nice to be able to view the rendered document in the interim.

index.js
```diff
@@ -0,0 +1,1044 @@
...
+                  #####                  #####
+                #####      ####################
+              #####       ######################
+            #####                     ###########
+          #####         ######################
+        #####          ####################
+      #####                        #####
+     ############                 ###
+    ############                 */
+
```
davidchambers: Could someone review the project's ASCII art? :grinning:

README.md
```diff
@@ -0,0 +1,576 @@
...
+methods which types must provide if they are to be considered members of
+these type classes. To be considered a [Functor][], for example, a type
+must provide a `fantasy-land/map` method which obeys the stated laws.
+
+This project provides four things:
+
+  - `TypeClass`, a function for defining arbitrary type classes;
+  - one `TypeClass` value for each Fantasy Land type class;
+  - lawful Fantasy Land methods for JavaScript's built-in types; and
+  - one function for each Fantasy Land method.
```
safareli: `toString`, `sequence`, `filter` and `filterM` don't fit in this 4 items. Maybe add a point saying that we also provide some derived methods and helpers?

bower.json
```diff
@@ -0,0 +1,35 @@
...
+  "keywords": [
+    "algebraic",
+    "functional",
+    "sanctuary",
+    "types"
+  ],
+  "dependencies": {},
+  "ignore": [
+    "/.git/",
+    "/bower_components/",
```
safareli: do we really need `bower_components` here?

README.md
davidchambers: Good catch! Updated.

package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "sanctuary-type-classes",
+  "version": "0.0.0",
+  "description": "Type classes for sanctuary-def",
```
davidchambers: This is out of date. The library is now useful on its own. How about `"Standard library for Fantasy Land"`?

safareli:
btw the table which was before showing which js object conforms to which FL type was pretty nice why was it removed?

davidchambers:
> btw the table which was before showing which js object conforms to which FL type was pretty nice why was it removed?

It couldn't handle `toString` and `fantasy-land/equals` because there are so many built-in types which can support these.

Before:

> | Type class      |   Array   | Function  |  Object   |  String   |
> |:--------------- |:---------:|:---------:|:---------:|:---------:|
> | [Semigroup][1]  |     ✔︎     |           |     ✔︎     |     ✔︎     |

After:

> `fantasy-land/concat` implementations are provided for the following built-in types: String, Array, and Object.

I'm happy to reinstate the table if you believe it better conveys the information, and if you can suggest a way to handle `toString` and `fantasy-land/equals`.


[1]: https://github.com/fantasyland/fantasy-land#semigroup

safareli:
> It couldn't handle `toString` and `fantasy-land/equals` because there are so many built-in types which can support these.

I did not get why it could not handle them?

If issue is that table looks bad because of too many types, then we can just combine those types which work only with `toString` and `equals` in one column as `Others *` and add a paragraph about which types are included in it

safareli:
But I cant wait to use this lib in [safareli/free#24](https://github.com/safareli/free/pull/24) so we can create that table in seperate PR and publish this first 🌯

Makefile
```diff
@@ -0,0 +1,80 @@
...
+README.md: index.js.tmp
+	$(TRANSCRIBE) \
+	  --heading-level 4 \
+	  --url 'https://github.com/sanctuary-js/sanctuary-type-classes/blob/v$(VERSION)/index.js#L{line}' \
+	  -- '$<' \
+	| LC_ALL=C sed 's/<h4 name="\(.*\)#\(.*\)">\(.*\)\1#\2/<h4 name="\1.prototype.\2">\3\1#\2/' >'$@'
+
+.INTERMEDIATE: index.js.tmp
+index.js.tmp: index.js
+	sed -e '/^[/][/]:/ s,\([A-Z][^ ]*\),<a href="#\1">\1</a>,g' -e 's,^//:,//.,' '$<' >'$@'
```
davidchambers: This line is neat. It turns all the names in the ASCII type-class graph into links to their respective sections. We get the benefits of fantasyland/fantasy-land#160 without adulterating the graph in __index.js__. :grinning:

safareli: Sweet 🍭

davidchambers:
I've spent the day sanding the patch. I'll merge this pull request tomorrow. You're welcome to take one last look before I do so. I'm particularly interested in knowing what people think of the [readme][1]. Having thought about and worked on this patch for several months I could easily have neglected to explain something important because I take it for granted. Let me know if this is the case.


[1]: https://github.com/sanctuary-js/sanctuary-type-classes/blob/dc-everything/README.md

kedashoe:
> I'm particularly interested in knowing what people think of the readme.

Looks good! A couple comments from someone who doesn't know much about this stuff :)

> one TypeClass value for each Fantasy Land type class;

I'm not sure exactly what this means.. A TypeClass "value"? I think I understand it after I got a little farther down, but maybe there is a more generic way to phrase this at the top so novices such as myself aren't thrown off? Maybe replace "value" with "definition" or just leave the word out altogether?

> one function for each Fantasy Land method;

Is this mostly a distinction between "function" and "method"? Is there actually a "Fantasy Land" instance that has methods?

- Something that doesn't need to be done now (or ever of course), but it might be cool for the TypeClass section to have a table that shows the result of the test for every built in type?

- The haskell-ish (I think?) examples from the function section are ok I think, although `empty(List) = Nil` made me pause for a bit and is maybe not necessary anyway?

davidchambers:
> A couple comments from someone who doesn't know much about this stuff

This is the most important perspective. :)

> > one TypeClass value for each Fantasy Land type class;
>
> I'm not sure exactly what this means.. A TypeClass "value"?

Let me explain the distinction I'm attempting to make, and perhaps you can suggest a better way to phrase it. I'm using _type class_ to refer to a group of types. JavaScript doesn't have types, so it doesn't have type classes, but the *idea* of type classes is still useful.

I'm using _`TypeClass` value_ as shorthand for _value of type `TypeClass`_. I'm referring to an actual JavaScript *value* which represents a particular type class.

> > one function for each Fantasy Land method;
>
> Is this mostly a distinction between "function" and "method"? Is there actually a "Fantasy Land" instance that has methods?

Again, it's possible I chose my words poorly. What I'm trying to say is that FL specifies the behaviour of `fantasy-land/map`, say, but rather than write `functor['fantasy-land/map'](f)` we can write `Z.map(f, functor)`. Not only is this convenient, but it allows `functor` to be `[1, 2, 3]` or some other value of a built-in type which *could* provide a `fantasy-land/map` method but does not.

I'm not sure what you mean by _a "Fantasy Land" instance that has methods_. Could you elaborate?

> Something that doesn't need to be done now (or ever of course), but it might be cool for the TypeClass section to have a table that shows the result of the test for every built in type?

I do like the idea. There is one [problem][1] we'll need to solve. Let's tackle it separately. :)

> The haskell-ish (I think?) examples from the function section are ok I think, although `empty(List) = Nil` made me pause for a bit and is maybe not necessary anyway?

I want to show that the function can operate on values of user-defined types. Does the confusion result from the fact that the List type is not described anywhere?


[1]: https://github.com/sanctuary-js/sanctuary-type-classes/pull/2#issuecomment-252511635

index.js
```diff
@@ -0,0 +1,1285 @@
...
+  var Function$prototype$ap = function(f) {
+    var apply = this;
+    return function(x) { return f(x)(apply(x)); };
+  };
+
+  //  Function$prototype$chain :: (a -> b) ~> (b -> a -> c) -> (a -> c)
+  var Function$prototype$chain = function(f) {
+    var chain = this;
+    return function(x) { return f(chain(x))(x); };
+  };
```
davidchambers: Fortuitously, @buzzdecafe opened ramda/ramda#1936 today which alerted me to the fact that Function can satisfy the requirements of Chain. The implementation is straightforward. In fact, it's almost identical to that of `Function$prototype$ap` since the only difference between the two functions is that one takes `(a -> b -> c)` whereas the other takes `(b -> a -> c)`.

kedashoe:
> I'm using TypeClass value as shorthand for value of type TypeClass. I'm referring to an actual JavaScript value which represents a particular type class.

I like this explanation :) And the [TypeClass](https://github.com/sanctuary-js/sanctuary-type-classes/blob/dc-everything/README.md#typeclass--stringarraytypeclassa-boolean---typeclass) also helps. Kind of just a problem the order the user sees the terms. Maybe a short glossary for terms like this that "value" could link to? Either way, not something to get hung up on at this point.

> Again, it's possible I chose my words poorly. What I'm trying to say is that FL...

Reading this one again I think what you have is good. The word "method" is used prominently in the FL spec, I just didn't realize that.

> I want to show that the function can operate on values of user-defined types. Does the confusion result from the fact that the List type is not described anywhere?

I see, I don't think I realized that (I like these examples more though knowing that!). Maybe a short sentence before these definitions to explain what `List` and `Cons` are?

index.js
davidchambers: I pushed a commit which includes the `'[object Object]'` check. Let me know what you think, @safareli. :)

index.js
davidchambers: The pull request now defines `Object#fantasy-land/reduce`.

index.js
davidchambers: This pull request now defines `filter` and `filterM`.

safareli:
🎉

PR #5, opened 2016-10-19
derive ‘lift2’ and ‘lift3’ from ‘map’ and ‘ap’

davidchambers:
Sanctuary provides [`S.lift`][1], [`S.lift2`][2], and [`S.lift3`][3]. `lift` is an alias for `map`, but it makes sense to define `lift2` and `lift3` here rather than in Sanctuary. `S.lift2` and `S.lift3` can then simply be curried equivalents of `Z.lift2` and `Z.lift3`.

[1]: https://sanctuary.js.org/#lift
[2]: https://sanctuary.js.org/#lift2
[3]: https://sanctuary.js.org/#lift3

### DIFF
--- a/index.js
+++ b/index.js
@@ -1037,6 +1037,46 @@
     return Apply.methods.ap(applyX)(applyF);
   };
 
+  //# lift2 :: Apply f => (a -> b -> c, f a, f b) -> f c
+  //.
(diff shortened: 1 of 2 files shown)


benperez:
Nice one.

:weight_lifting_woman:

safareli:
👍  nice examples

PR #12, opened 2016-11-05
fantasy-land@2.1.0

davidchambers:
This pull request adds `TypeClass` values for [Alt][1], [Plus][2], and [Alternative][3], and functions for [`fantasy-land/alt`][4] and [`fantasy-land/zero`][5], all of which were added in fantasyland/fantasy-land#197.

Please review this when you have some time, @safareli. :)

[1]: https://github.com/fantasyland/fantasy-land#alt
[2]: https://github.com/fantasyland/fantasy-land#plus
[3]: https://github.com/fantasyland/fantasy-land#alternative
[4]: https://github.com/fantasyland/fantasy-land#alt-method
[5]: https://github.com/fantasyland/fantasy-land#zero-method

### DIFF
--- a/index.js
+++ b/index.js
@@ -31,34 +31,37 @@
 //. ## Type-class hierarchy
 //.
 //. <pre>
-//:  Setoid   Semigroup   Foldable           Functor            Extend
-//: (equals)   (concat)   (reduce)            (map)            (extend)
-//:               |           \             / | | | \             /
-//:               |            \           /  | | |  \           /
-//:               |             \         /   | | |   \         /
-//:               |              \       /    | | |    \       /
-//:               |               \     /     | | |     \     /
-//:               |                \   /      | | |      \   /
-//:               |                 \ /       / | \       \ /
-//:            Monoid           Traversable  /  |  \    Comonad
-//:            (empty)          (traverse)  /   |   \  (extract)
-//:                                        /    |    \
-//:                               Bifunctor   Apply   Profunctor
-//:                                (bimap)     (ap)    (promap)
-//:                                            / \
-//:                                           /   \
-//:                                          /     \
-//:                                         /       \
-//:                                        /         \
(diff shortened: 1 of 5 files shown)


safareli:
If Array is alternative String could be Alternative in the same way too.

list/Array/String we could have other instance, which just does concatenation. In [purescript][ps] and [GHC][ghc] (search `Alternative []`) we have that concatenative version so maybe it's better to stay consistent with them?

I could not find implementation of alternative for Map nor in PS nor in Haskell , but If we follow use concatenation as `alt` for string and array maybe we should do same for `StrMap` ?


[ghc]: https://hackage.haskell.org/package/base-4.9.0.0/docs/src/GHC.Base.html#line-756
[ps]: https://github.com/purescript/purescript-control/blob/master/src/Control/Alt.purs#L26-L27

davidchambers:
Thanks for the feedback, @safareli. I've updated the patch. Could you take another look?

> If Array is alternative String could be Alternative in the same way too.

Can it, though? String does not satisfy the requirements of Functor.

safareli:
> Can it, though? String does not satisfy the requirements of Functor.

oh, yes :d

index.js
```diff
@@ -353,6 +356,45 @@
...
+  //# Alternative :: TypeClass
+  //.
+  //. `TypeClass` value for [Alternative][].
+  //.
+  //. ```javascript
+  //. > Alternative.test([])
+  //. true
+  //.
+  //. > Alternative.test({})
+  //. false
```
safareli: I think it should be true

davidchambers: I don't think so. We can't define `fantasy-land/of` for Object, so Object can't satisfy Applicative.

safareli: oh, yes, again

safareli:
LGTM

PR #17, opened 2016-11-16
recognize equivalent circular data structures in ‘equals’

davidchambers:
Many thanks to @vyznev for suggesting this approach on [Stack Overflow][1].

@Avaq, this is the part where you try to break it. :)

[1]: http://stackoverflow.com/a/40622794/312785

### DIFF
--- a/index.js
+++ b/index.js
@@ -497,7 +497,7 @@
   var Number$prototype$equals = function(other) {
     return typeof other === 'object' ?
       typeof this === 'object' &&
-        equals(other.valueOf(), this.valueOf()) :
+        equals(this.valueOf(), other.valueOf()) :
       isNaN(other) && isNaN(this) ||
         other === this && 1 / other === 1 / this;
   };
@@ -510,7 +510,7 @@
 
   //  Date$prototype$equals :: Date ~> Date -> Boolean
   var Date$prototype$equals = function(other) {
-    return equals(other.valueOf(), this.valueOf());
+    return equals(this.valueOf(), other.valueOf());
   };
 
   //  RegExp$prototype$equals :: RegExp ~> RegExp -> Boolean
@@ -602,7 +602,7 @@
   var Array$prototype$equals = function(other) {
     if (other.length !== this.length) return false;
     for (var idx = 0; idx < this.length; idx += 1) {
-      if (!equals(other[idx], this[idx])) return false;
+      if (!equals(this[idx], other[idx])) return false;
     }
     return true;
   };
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -602,7 +602,7 @@
   var Array$prototype$equals = function(other) {
     if (other.length !== this.length) return false;
     for (var idx = 0; idx < this.length; idx += 1) {
-      if (!equals(other[idx], this[idx])) return false;
+      if (!equals(this[idx], other[idx])) return false;
```
davidchambers: In order to compare `[1, <Circular>]` and `[1, [1, <Circular>]]` we expect to compare `<Circular>` and `[1, <Circular>]` (after comparing `1` and `1`), but with the previous argument order the values would be flipped: `[1, <Circular>]` and `<Circular>`. The change above isn't necessary, but avoids confusion when debugging.

index.js
```diff
@@ -738,6 +738,11 @@
     return function(_) { return x; };
   };
 
+  //  Function$prototype$equals :: Function ~> Function -> Boolean
+  var Function$prototype$equals = function(other) {
+    return other === this;
+  };
+
```
davidchambers: Having removed `Object(x) === Object(y)` from `equals` we need to define `Function#fantasy-land/equals`.

test/index.js
```diff
@@ -383,7 +386,8 @@ test('toString', function() {
   eq(Z.toString(['foo', 'bar']), '["foo", "bar"]');
   eq(Z.toString(/x/.exec('xyz')), '["x", "index": 0, "input": "xyz"]');
   eq(Z.toString((function() { var xs = []; xs.z = true; xs.a = true; return xs; }())), '["a": true, "z": true]');
-  eq(Z.toString(circular), '["foo", <Circular>]');
+  eq(Z.toString(ones), '[1, <Circular>]');
+  eq(Z.toString(ones_), '[1, [1, <Circular>]]');
```
davidchambers: Having two string representations of one value is somewhat disturbing, but I don't see an alternative.

index.js
```diff
@@ -932,25 +944,27 @@
...
-    //  $seen :: Array Any
-    var $seen = [];
+    //  $pairs :: Array (Pair Any Any)
+    var $pairs = [];
 
-    //  equal :: (a, b) -> Boolean
-    var equal = function(x, y) {
-      $seen.push(x, y);
+    return function equals(x, y) {
+      if (type(x) !== type(y)) {
```
rjmk: Incredible how much of the solution Sanctuary's typechecking captures!

davidchambers: I :heart: the simplicity of the approach. _Are you two members of the same type? Is that type a setoid? Okay, we'll let its `fantasy-land/equals` method determine whether you are equal._

Issue #18, opened 2016-11-22
traverse not stack safe


syaiful6:
title say it all, small repro:

https://jsfiddle.net/syaiful6/reh4dwnd/

davidchambers:
Is it specifically [`Array$prototype$traverse`][1] that's problematic, do you think? If so, are you interested in working on an implementation which does not blow the stack? I'm not sure that it's possible, but it's definitely worth investigating. :)


[1]: https://github.com/sanctuary-js/sanctuary-type-classes/blob/v1.2.0/index.js#L646-L653

syaiful6:
yes, it is. it's happen because large ```ap``` operation on there. I have an implementation that doesn't blow the stack. I will pull a PR when i have time.

syaiful6:
okay i already open a PR, ready to review.

PR #19, opened 2016-11-22
(fix traverse) make traverse stack safe

syaiful6:
close https://github.com/sanctuary-js/sanctuary-type-classes/issues/18

### DIFF
--- a/index.js
+++ b/index.js
@@ -83,9 +83,11 @@
 
   'use strict';
 
-  //  has :: (String, Object) -> Boolean
-  function has(k, o) {
-    return Object.prototype.hasOwnProperty.call(o, k);
+  //  concat_ :: Array a -> Array a -> Array a
+  function concat_(xs) {
+    return function(ys) {
+      return xs.concat(ys);
+    };
   }
 
   //  constant :: a -> b -> a
@@ -95,13 +97,18 @@
     };
   }
 
+  //  has :: (String, Object) -> Boolean
+  function has(k, o) {
+    return Object.prototype.hasOwnProperty.call(o, k);
+  }
+
   //  identity :: a -> a
   function identity(x) { return x; }
 
-  //  prepend :: a -> Array a -> Array a
-  function prepend(x) {
-    return function(xs) {
-      return [x].concat(xs);
+  //  pair :: a -> b -> Pair a b
+  function pair(x) {
+    return function(y) {
+      return [x, y];
     };
   }
 
@@ -651,11 +658,19 @@
 
   //  Array$prototype$traverse :: Applicative f => Array a ~> (a -> f b, c -> f c) -> f (Array b)
   function Array$prototype$traverse(f, of) {
-    var applicative = of([]);
-    for (var idx = this.length - 1; idx >= 0; idx -= 1) {
-      applicative = ap(map(prepend, f(this[idx])), applicative);
+    var xs = this;
+    function go(idx, n) {
+      switch (n) {
+        case 0: return of([]);
+        case 2: return lift2(pair, f(xs[idx]), f(xs[idx + 1]));
+        default:
+          var m = Math.floor(n / 4) * 2;
+          return lift2(concat_, go(idx, m), go(idx + m, n - m));
+      }
     }
-    return applicative;
+    return this.length % 2 === 1 ?
+      lift2(concat_, map(Array$of, f(this[0])), go(1, this.length - 1)) :
+      go(0, this.length);
   }
 
   //  Array$prototype$extend :: Array a ~> (Array a -> b) -> Array b
--- a/test/Lazy.js
+++ b/test/Lazy.js
@@ -1,0 +1,29 @@
+'use strict';
+
+var FL = require('fantasy-land');
+
+var Z = require('..');
+
+
+//  Lazy :: (() -> a) -> Lazy a
+function Lazy(f) {
+  if (!(this instanceof Lazy)) return new Lazy(f);
+  this.run = f;
+}
+
+Lazy[FL.of] = function(a) {
+  return Lazy(function() { return a; });
+};
+
+Lazy.prototype['@@type'] = 'sanctuary-type-classes/Lazy';
+
+Lazy.prototype[FL.map] = function(f) {
+  return Z.ap(Z.of(Lazy, f), this);
+};
+
+Lazy.prototype[FL.ap] = function(other) {
+  var task = this;
+  return Lazy(function() { return other.run()(task.run()); });
+};
+
+module.exports = Lazy;
--- a/test/index.js
+++ b/test/index.js
@@ -6,6 +6,7 @@
 var Z = require('..');
 
 var Identity = require('./Identity');
+var Lazy = require('./Lazy');
 var List = require('./List');
 var Maybe = require('./Maybe');
 var Tuple = require('./Tuple');
@@ -23,6 +24,12 @@ var Just = Maybe.Just;
 function Array$of(x) {
   eq(arguments.length, Array$of.length);
   return Z.of(Array, x);
+}
+
+//  Lazy$of :: a -> Lazy a
+function Lazy$of(x) {
+  eq(arguments.length, Lazy$of.length);
+  return Z.of(Lazy, x);
 }
 
 //  List$of :: a -> List a
@@ -772,6 +779,7 @@ test('traverse', function() {
   eq(Z.traverse(Identity, identInc, [1, 2, 3]), Identity([2, 3, 4]));
   eq(Z.traverse(Identity, identInc, Nil), Identity(Nil));
   eq(Z.traverse(Identity, identInc, Cons(1, Cons(2, Cons(3, Nil)))), Identity(Cons(2, Cons(3, Cons(4, Nil)))));
+  eq(Z.traverse(Lazy$of, Lazy$of, range(70000)).run().length, 70000);
 });
 
 test('sequence', function() {
--- a/Makefile
+++ b/Makefile
@@ -78,6 +78,6 @@ setup:
 
 .PHONY: test
 test:
-	$(ISTANBUL) cover node_modules/.bin/_mocha -- --timeout 10000 --ui tdd -- test/index.js
+	$(ISTANBUL) cover node_modules/.bin/_mocha -- --timeout 30000 --ui tdd -- test/index.js
 	$(ISTANBUL) check-coverage --branches 100
 	make doctest


syaiful6:
there are error on linting, i don't know if i should delete ```prepend``` or not, if it should deleted let me know it.

safareli:
`Z.traverse(Identity, identInc, range(70000))` using current implementation is producing correct result without any overflow. can you update test such that current implementation fails with overflow?

syaiful6:
@safareli okay, updated still passed!

safareli:
what I was saying is that, the test case you have added works even with the implementation we have currently:
```
//  Array$prototype$traverse :: Applicative f => Array a ~> (a -> f b, c -> f c) -> f (Array b)
var Array$prototype$traverse = function(f, of) {
  var applicative = of([]);
  for (var idx = this.length - 1; idx >= 0; idx -= 1) {
    applicative = ap(map(prepend, f(this[idx])), applicative);
  }
  return applicative;
};
// this works ok
Z.traverse(Identity, identInc, range(70000)
```

syaiful6:
Its because identity ap operation that make it safe for current implementation. Try with my issue version StateR applicative will fail. Please see my issue link on fiddle.

syaiful6:
Oh I see. Should I put StateR applicative there?

safareli:
yes i think it would be good to add minimal implementation of StateR such that tests fail using current traverse

syaiful6:
@safareli thanks! updated and added test that will fail for current implementation but success for this one.

Note: i also delete prepend because it's no longer use.

davidchambers:
Could you explain how the algorithm works, @syaiful6? I'd like to understand this part in particular:

```javascript
var pivot = bottom + Math.floor((top - bottom) / 4) * 2;
return lift2(concat2, go.call(this, bottom, pivot),
             go.call(this, pivot, top));
```

syaiful6:
@davidchambers it's using top-down [divide and conquer algorithm](https://en.wikipedia.org/wiki/Divide_and_conquer_algorithms). Here, we traverse the halves then ```lift2```. it works because ```ap``` is associative. Stack consumption while traversing is logarithmic.

Basically the pivot calculation decide how to divide the array, the idea is simply to choose pivots that will yield two even-length partitions. Because for array with 6 elements, a pivot of 2 is superior to a pivot of 3. 

Note: the current implementation and this version, will success to generate applicative action, eg calling ```travese``` will not cause stack overflow for both, but the current implementation will fail when the action actually run.

Given an array length ```500``` means the current implementation will gives a bunch of repeated frames from the ```ap``` or ```chain``` if it derived from it on underlying Applicative, to more precise it will consumes 1001 frames, but in this implementation only consume 24.

So, this version basically helps prevent stack overflows arising from large numbers of chained ```ap```s. Not when it build the action itself. Thay why i do ```Z.traverse(StateL[FL.of], subAccum, range(70000))._f(100)``` rather than simply ```Z.traverse(StateL[FL.of], subAccum, range(70000))```.

Further ref about how we able build the array on the underlying Applicative: https://pdfs.semanticscholar.org/461a/48f6fd6951a22eee00578e25b7af47b14a3d.pdf

davidchambers:
Terrific explanation, @syaiful6! Thank you for taking the time to write it.

index.js
```diff
@@ -644,13 +637,44 @@
...
+          case 2: return lift2(array2, f(this[bottom]), f(this[bottom + 1]));
+          case 3: return lift3(array3, f(this[bottom]), f(this[bottom + 1]),
+                               f(this[bottom + 2]));
+          default:
+            var pivot = bottom + Math.floor((top - bottom) / 4) * 2;
+            return lift2(concat2, go.call(this, bottom, pivot),
+                         go.call(this, pivot, top));
+        }
+      };
+      return go.call(this, 0, this.length);
```
davidchambers: Is there a reason to provide the array as an implicit argument? Could we use `go(this, 0, this.length)` instead?

syaiful6: not really @davidchambers, i just want to keep the this ```context```

index.js
```diff
@@ -644,13 +637,44 @@
...
+        switch (top - bottom) {
+          case 0: return of([]);
+          case 1: return map(array1, f(this[bottom]));
+          case 2: return lift2(array2, f(this[bottom]), f(this[bottom + 1]));
+          case 3: return lift3(array3, f(this[bottom]), f(this[bottom + 1]),
+                               f(this[bottom + 2]));
+          default:
+            var pivot = bottom + Math.floor((top - bottom) / 4) * 2;
+            return lift2(concat2, go.call(this, bottom, pivot),
+                         go.call(this, pivot, top));
```
syaiful6: ok, that make sense.

index.js
```diff
@@ -644,13 +637,44 @@
...
+      };
+    };
+    return function(f, of) {
+      var go = function(bottom, top) {
+        switch (top - bottom) {
+          case 0: return of([]);
+          case 1: return map(array1, f(this[bottom]));
+          case 2: return lift2(array2, f(this[bottom]), f(this[bottom + 1]));
+          case 3: return lift3(array3, f(this[bottom]), f(this[bottom + 1]),
+                               f(this[bottom + 2]));
```
davidchambers: I'm trying to understand the relationship between these cases and the default case. Which of the cases above are base cases, and which are included solely to reduce stack usage?

index.js
davidchambers: I'd prefer to use `go(this, 0, this.length)` and to replace the implicit `this` with an explicit `xs` within `go`.

syaiful6:
@davidchambers okay, updated

index.js
syaiful6: good idea! @davidchambers

safareli:
This optimisation looks good. 

For a bit I was thinking if issue was in `StateL` itself, I took a look at cats/[StateT](https://github.com/typelevel/cats/blob/c56fbaba91dbce976e933b5f19990cdd7b604737/core/src/main/scala/cats/data/StateT.scala), it has test ["traversing is being stack-safe"](https://github.com/typelevel/cats/blob/c56fbaba91dbce976e933b5f19990cdd7b604737/tests/src/test/scala/cats/tests/StateTTests.scala#L18-L22), then to [traverse of List](https://github.com/typelevel/cats/blob/f90245dbec3d683f2bdf57d63cab8784d6d04115/core/src/main/scala/cats/instances/list.scala#L68-L71) which at first looks like the implementation we had before, but It uses `map2Eval` and foldRight which are using some [Eval](http://typelevel.org/cats/api/cats/Eval.html) structure which look slike alows List to be lazy and not cause stack issues, and then I stoped :D

test/index.js
```diff
@@ -147,6 +152,14 @@ var wrap = function(before) {
   };
 };
 
+var subAccum = function(a) {
```
safareli: this is used once should we inline it?

safareli: this is used once, should we inline it?

test/index.js
```diff
@@ -764,6 +777,11 @@ test('traverse', function() {
   eq(Z.traverse(Identity, identInc, [1, 2, 3]), Identity([2, 3, 4]));
   eq(Z.traverse(Identity, identInc, Nil), Identity(Nil));
   eq(Z.traverse(Identity, identInc, Cons(1, Cons(2, Cons(3, Nil)))), Identity(Cons(2, Cons(3, Cons(4, Nil)))));
+
+  assert.doesNotThrow(function() {
+    // traverse and run the accumulator. this should stack safe.
+    Z.traverse(StateL[FL.of], subAccum, range(70000))._f(100);
+  }, RangeError);
```
davidchambers: The documentation is somewhat confusing, but it seems we needn't provide this argument.

test/index.js
```diff
@@ -764,6 +777,11 @@ test('traverse', function() {
   eq(Z.traverse(Identity, identInc, [1, 2, 3]), Identity([2, 3, 4]));
   eq(Z.traverse(Identity, identInc, Nil), Identity(Nil));
   eq(Z.traverse(Identity, identInc, Cons(1, Cons(2, Cons(3, Nil)))), Identity(Cons(2, Cons(3, Cons(4, Nil)))));
+
+  assert.doesNotThrow(function() {
+    // traverse and run the accumulator. this should stack safe.
```
davidchambers: Let's begin each sentence with a capital letter:

```javascript
    //  Traverse and run the accumulator. This should be stack safe.
```

test/index.js
davidchambers: I don't mind this being defined separately, but I would like to see it defined closer to its call site (somewhere in the `test('traverse', ...)` block).

test/StateL.js
```diff
@@ -0,0 +1,47 @@
...
+  return StateL(function(s) {
+    return {
+      accum: s,
+      value: x
+    };
+  });
+};
+
+StateL.prototype[FL.map] = function(f) {
+  var fn = this._f;
```
davidchambers: It's a bit awkward to have two functions named `f` and `fn`. It reminds me a bit of `String#substr` and `String#substring`, something which has bothered me for some time. Could we name this `_f` instead? At least then we'd be mirroring the field name.

index.js
```diff
@@ -98,10 +98,17 @@
...
-  //  prepend :: a -> Array a -> Array a
-  var prepend = function(x) {
-    return function(xs) {
-      return [x].concat(xs);
+  //  concat_ :: Array a -> Array a -> Array a
+  var concat_ = function(xs) {
+    return function(ys) {
+      return xs.concat(ys);
+    };
+  };
```
davidchambers: Let's preserve the alphabetical ordering of our helper functions. :)

test/index.js
safareli: And as I understand it's not quite stack safe, just usage of stack is logarithmic. so I think it would  be better to be much precise here, also it would be nice to state this in description of the Array.traverse.

davidchambers:
What's the status of this pull request, @syaiful6? I'm happy to take it over if you're busy with other things.

index.js
syaiful6: well, i don't know if it alpha ordering there, if it should ```has``` must be below ```constant``` function. so should i change it too?

test/StateL.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+var FL = require('fantasy-land');
+
+
+var StateL = function StateL(f) {
```
syaiful6: Hmm, not to confuse it with ```State``` monad, it just applicative that accumulate it results from ```left```, so ```L``` there stand for ```Left```

test/index.js
syaiful6: what's does mean? Do you think we don't need provide ```RangeError``` there?

test/index.js
```diff
@@ -764,6 +777,11 @@ test('traverse', function() {
...
+    Z.traverse(StateL[FL.of], subAccum, range(70000))._f(100);
```
syaiful6: @safareli not easy to check it! the results is an object with key accum and value, the value one contains array length ```70000```, and the member subtracted (100) from the left, so it
```
[0 - 100, 1 - (0 - 100), 2 - (1 - (0 -100)), 3 - 2 - (1 - (0 -100)), ... ]
```
if you know how to tests it, i will happy to add it.

test/index.js
davidchambers: That's right. `assert.doesNotThrow(<function>)` should work.

index.js
davidchambers: Yes, please do. :)

safareli:
Actually the StateL is not stack safe itself, not the traverse. 

The `map` and `ap` of the structure re not stack safe as they are building up computation lazily, on each call they create a function which is using passed in function requiring larger stack size, leading to stack overflow (when you run `._f`)

`.  apply curried function `f` which is taking 10000 arguments:
```js
...(StateL.of(9999).ap(StateL.of(10000).ap(StateL.of(f)))
```
2. map on some StateL 10000 times
```
StateL.of(1).map(id).map(id).map(id).map(id).map(id).....
``` 

So the issue is with all lazy structure `data.task`, `fantasy-state` ... as all of them need extra stack frame on each `map/ap/chain`.

P.S. In case you missed this is how i fixed stack issue with long chain of [function composition](https://medium.com/@safareli/stack-safe-function-composition-85d61feee37e).

syaiful6:
okay updated, a bit trouble one, look like the coding style changed.

safareli:
👍

davidchambers:
I will see whether I have write access to your branch, @syaiful6. I would like to rebase it to avoid an extra merge commit in the history. If I have write access I will do this myself to save you the trouble.

test/Lazy.js
```diff
@@ -0,0 +1,28 @@
...
+}
+
+Lazy[FL.of] = function(a) {
+  return Lazy(function() { return a; });
+};
+
+Lazy.prototype['@@type'] = 'sanctuary-type-classes/Lazy';
+
+Lazy.prototype[FL.map] = function(f) {
+  return this[FL.ap](Lazy[FL.of](f));
```
davidchambers: I think it would be nice to use `Z.ap` and `Z.of` here:

```javascript
  return Z.ap(Z.of(Lazy, f), this);
```

test/Lazy.js
```diff
@@ -0,0 +1,28 @@
...
+};
+
+Lazy.prototype[FL.ap] = function(other) {
+  var f = other.run;
+  var v = this.run;
+  return Lazy(function() { return f()(v()); });
```
davidchambers: What do you think of this change?

```javascript
  var lazy = this;
  return Lazy(function() { return other.run()(lazy.run()); });
```

I like to avoid variables where possible. Perhaps you wanted to avoid accessing the `run` property multiple times, though?

test/index.js
```diff
@@ -764,6 +771,7 @@ test('traverse', function() {
   eq(Z.traverse(Identity, identInc, [1, 2, 3]), Identity([2, 3, 4]));
   eq(Z.traverse(Identity, identInc, Nil), Identity(Nil));
   eq(Z.traverse(Identity, identInc, Cons(1, Cons(2, Cons(3, Nil)))), Identity(Cons(2, Cons(3, Cons(4, Nil)))));
+  eq(Z.traverse(Lazy$of, Lazy$of, range(70000)).run(), range(70000));
```
safareli:
```js
function computeCallStack() {
  try {
    return 1 + computeCallStack()
  } catch (_) {
    return 1
  }
}

const MAX_STACK = computeCallStack()
```
stack sizes [Node: 11034, Firefox: 50994, Chrome: 10402](http://www.2ality.com/2014/04/call-stack-size.html)

In node it's size is around 10k so if we use `MAX_STACK` instead of 70k, timeout issue might go away

test/Lazy.js
safareli: it's fine.

test/index.js
syaiful6: not sure, the ```eq``` check make it more expensive, because it  check equality based ```Z.toString``` and ```assertEquals``` so we loop 2 big array twice here. I dont see the need for equality check, because what we really care are the stack overflow. And the correctness of ```Z.traverse``` already checked on other tests. That why the previous test are just execute the applicative action without equality check.

safareli:
we can just check lenghh and last element for example

davidchambers:
For some reason CircleCI is not running the tests for this pull request. Does anyone have experience with resolving this problem?

davidchambers:
I managed to "trick" CircleCI into running the tests by pushing to a `stack-safe-traverse` branch in *this* repository (which shouldn't have had any effect). @safareli, please confirm that you're happy with this pull request in its current state. Are you happy with the tweaks I made, @syaiful6?

safareli:
👍

davidchambers:
Thank you for your contribution, @syaiful6. This is a nice improvement. :thumbsup:

PR #22, opened 2016-12-09
treat prototype objects as plain objects

davidchambers:
Commit message:

> The prototype object associated with some type T is not a member of T, even if its `@@type` property indicates that it is. The object may be missing fields set by T's data constructors. The Left and Right data constructors of the Either type, for example, could define `isLeft`, `isRight`, and `value` fields which various methods assume to exist. The following method invocation is therefore incorrect:
>
>     Either.prototype['fantasy-land/equals'](x)
>
> This commit adds `x.constructor.prototype !== x` checks to `type` and `getBoundMethod` so that prototype objects are treated as plain objects.
>
> `@@type` properties should now be defined on type representatives rather than on prototype objects. This is a breaking change.

This issue took many hours to understand and fix. I'd like a couple of people to review this patch before we merge it. :)

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -3,6 +3,8 @@
 /* eslint-env mocha */
 /* eslint max-len: "off" */
 
+var type = require('sanctuary-type-identifiers');
+
 var Z = require('..');
 
 var Identity = require('./Identity');
@@ -172,7 +174,7 @@ test('TypeClass', function() {
   //  Bar :: TypeClass
   var Bar = Z.TypeClass('my-package/Bar', [Foo], hasMethod('bar'));
 
-  eq(Foo['@@type'], 'sanctuary-type-classes/TypeClass');
+  eq(type(Foo), 'sanctuary-type-classes/TypeClass');
   eq(Foo.name, 'my-package/Foo');
   eq(Foo.test(null), false);
   eq(Foo.test({}), false);
@@ -180,7 +182,7 @@ test('TypeClass', function() {
   eq(Foo.test({bar: function() {}}), false);
   eq(Foo.test({foo: function() {}, bar: function() {}}), true);
 
-  eq(Bar['@@type'], 'sanctuary-type-classes/TypeClass');
+  eq(type(Bar), 'sanctuary-type-classes/TypeClass');
   eq(Bar.name, 'my-package/Bar');
   eq(Bar.test(null), false);
   eq(Bar.test({}), false);
@@ -190,17 +192,17 @@ test('TypeClass', function() {
 });
 
 test('Setoid', function() {
-  eq(Z.Setoid['@@type'], 'sanctuary-type-classes/TypeClass');
+  eq(type(Z.Setoid), 'sanctuary-type-classes/TypeClass');
   eq(Z.Setoid.name, 'sanctuary-type-classes/Setoid');
   eq(Z.Setoid.test(null), true);
   eq(Z.Setoid.test(''), true);
   eq(Z.Setoid.test([]), true);
   eq(Z.Setoid.test({}), true);
-  eq(Z.Setoid.test({'@@type': 'my-package/Quux'}), false);
+  eq(Z.Setoid.test({'@@type': 'my-package/Quux'}), true);
 });
 
 test('Semigroup', function() {
-  eq(Z.Semigroup['@@type'], 'sanctuary-type-classes/TypeClass');
+  eq(type(Z.Semigroup), 'sanctuary-type-classes/TypeClass');
   eq(Z.Semigroup.name, 'sanctuary-type-classes/Semigroup');
   eq(Z.Semigroup.test(null), false);
   eq(Z.Semigroup.test(''), true);
@@ -209,7 +211,7 @@ test('Semigroup', function() {
 });
 
 test('Monoid', function() {
-  eq(Z.Monoid['@@type'], 'sanctuary-type-classes/TypeClass');
+  eq(type(Z.Monoid), 'sanctuary-type-classes/TypeClass');
   eq(Z.Monoid.name, 'sanctuary-type-classes/Monoid');
   eq(Z.Monoid.test(null), false);
   eq(Z.Monoid.test(''), true);
@@ -218,7 +220,7 @@ test('Monoid', function() {
 });
 
 test('Functor', function() {
-  eq(Z.Functor['@@type'], 'sanctuary-type-classes/TypeClass');
+  eq(type(Z.Functor), 'sanctuary-type-classes/TypeClass');
   eq(Z.Functor.name, 'sanctuary-type-classes/Functor');
   eq(Z.Functor.test(null), false);
   eq(Z.Functor.test(''), false);
@@ -227,7 +229,7 @@ test('Functor', function() {
 });
 
 test('Bifunctor', function() {
-  eq(Z.Bifunctor['@@type'], 'sanctuary-type-classes/TypeClass');
+  eq(type(Z.Bifunctor), 'sanctuary-type-classes/TypeClass');
   eq(Z.Bifunctor.name, 'sanctuary-type-classes/Bifunctor');
   eq(Z.Bifunctor.test(null), false);
   eq(Z.Bifunctor.test(''), false);
@@ -237,7 +239,7 @@ test('Bifunctor', function() {
 });
 
 test('Profunctor', function() {
-  eq(Z.Profunctor['@@type'], 'sanctuary-type-classes/TypeClass');
+  eq(type(Z.Profunctor), 'sanctuary-type-classes/TypeClass');
   eq(Z.Profunctor.name, 'sanctuary-type-classes/Profunctor');
   eq(Z.Profunctor.test(null), false);
   eq(Z.Profunctor.test(''), false);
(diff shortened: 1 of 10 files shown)


test/index.js
```diff
@@ -189,7 +189,7 @@ test('Setoid', function() {
   eq(Z.Setoid.test(''), true);
   eq(Z.Setoid.test([]), true);
   eq(Z.Setoid.test({}), true);
-  eq(Z.Setoid.test({'@@type': 'my-package/Quux'}), false);
+  eq(Z.Setoid.test({'@@type': 'my-package/Quux'}), true);
```
safareli: So this means the object is Setoid as it's a normal object, which is setoid, and the fact that it has `@@type` as it's own prop does not makes it a different object ?

davidchambers: That's right. `{'@@type': 'my-package/Quux'}` is considered the prototype for the Quux type, so its `@@type` property is ignored.

test/index.js
```diff
@@ -513,7 +513,9 @@ test('equals', function() {
   eq(Z.equals(Math.sin, Math.cos), false);
   eq(Z.equals(Identity(Identity(Identity(0))), Identity(Identity(Identity(0)))), true);
   eq(Z.equals(Identity(Identity(Identity(0))), Identity(Identity(Identity(1)))), false);
-  eq(Z.equals({'@@type': 'my-package/Quux'}, {'@@type': 'my-package/Quux'}), false);
+  eq(Z.equals({'@@type': 'my-package/Quux'}, {'@@type': 'my-package/Quux'}), true);
```
safareli: so here `Z.equals` delegates to `equals` from Setoid instance of Object even if this objects had `fl.equals` defined

davidchambers: That's right. Note how `Z.equals` behaves on `master` when giving a prototype object:

```javascript
> var Z = require('.')
undefined
> var Maybe = require('./test/Maybe')
undefined
> Maybe.prototype
Maybe {
  '@@type': 'sanctuary-type-classes/Maybe',
  'fantasy-land/equals': [Function],
  'fantasy-land/map': [Function],
  'fantasy-land/alt': [Function],
  toString: [Function],
  inspect: [Function] }
> Z.equals(Maybe.prototype, Maybe.prototype)
undefined
```

How does `Z.equals(Maybe.prototype, Maybe.prototype)` evaluate to `undefined`?! By dispatching to `Maybe.prototype['fantasy-land/equals']` **with `Maybe.prototype` as the implicit argument**. Here's the method's definition:

```javascript
Maybe.prototype[FL.equals] = function(other) {
  return this.isNothing ? other.isNothing
                        : other.isJust && Z.equals(other.value, this.value);
};
```

We essentially perform this method invocation:

```javascript
Maybe.prototype['fantasy-land/equals'](Maybe.prototype)
```

Substituting `Maybe.prototype` for `this` and `other`:

```javascript
Maybe.prototype.isNothing ?
  Maybe.prototype.isNothing :
  Maybe.prototype.isJust && Z.equals(Maybe.prototype.value, Maybe.prototype.value)
```

Evaluating the condition:

```javascript
undefined ?
  Maybe.prototype.isNothing :
  Maybe.prototype.isJust && Z.equals(Maybe.prototype.value, Maybe.prototype.value)
```

Taking the appropriate branch:

```javascript
Maybe.prototype.isJust && Z.equals(Maybe.prototype.value, Maybe.prototype.value)
```

Evaluating the left-hand operand:

```javascript
undefined && Z.equals(Maybe.prototype.value, Maybe.prototype.value)
```

Short circuiting:

```javascript
undefined
```

This makes it clear that methods should not receive a prototype object as `this`.

Avaq:
:eyes:

Avaq:
I consider this to be quite the drawback. I think this forces people to define types through `prototype`:

Let's define a simple Future-like Functor type without using Prototype (plain object factory):

```js
const Async = run => ({
  run,
  '@@type': 'My/Async',
  'fantasy-land/map': f => Async(g => run(x => g(f(x))))
});
Async.of = x => Async(f => f(x));
```

And the following code run on `sanctuary-type-classes@master` produces an expected result:

```js
> Z.map(x => x + 1, Async.of(1)).run(console.log);
2
```

However, when run on this branch:

```js
> Z.map(x => x + 1, Async.of(1)).run(console.log);
TypeError: Z.map(...).run is not a function

> Z.map(x => x + 1, Async.of(1))
{ run: 'f => f(x)1',
  '@@type': 'My/Async1',
  'fantasy-land/map': 'f => Async(g => run(x => g(f(x))))1' }
```

The last example shows that `Z.map` is now dispatching to `Object$prototype$map`, concatenating `1` at the end of every property on the "object".

I should add that behavior remains the same for objects which do not define a `@@type` property. So if we had defined our `Async` type like most people do (without the property), behavior would not have changed. I'm not sure how many people within the community actually own types which are defined using plain object factories *and* have a `@@type` property.. possibly none.

davidchambers:
@safareli asked on [Gitter][1] why `@@type` is defined on the prototype object rather than on the type representative. I replied:

> Initially, @safareli, there was no requirement that `@@type` be defined on a value's prototype object specifically. If `x['@@type']` evaluated to `'foo/Foo'`, `x` would be considered a member of the Foo type regardless of whether `x` had an own `@@type` property or whether it inherited this property from its prototype. This left the internal structure of ADTs unspecified for the benefit of ADT library authors.

[`S.is`][2] requires a way to get a type identifier from a type representative. This is impossible if a type identifier is only available as a property of members of the type. So the type identifier needs to be accessible from both directions: from the type representative, and from members. There are two reasonable locations for the `@@type` property:

1.  Prototype object
      - `typeRep.prototype['@@type']`
      - `member['@@type']` (or `member.constructor.prototype['@@type']`)

2.  Type representative
      - `typeRep['@@type']`
      - `member.constructor['@@type']`

Initially I could see no reason to favour one location over the other. Prompted by @Avaq's example above, I now see a good reason to define `@@type` on the type representative: it's trivial to define a `constructor` property when using a "factory" function, but wiring the prototype chain is far from trivial.

I've pushed a second commit which moves the `@@type` property from the prototype object to the type representative. The Async example above can now be written like so:

```javascript
function Async(run) {
  return {
    constructor: Async,
    run: run,
    'fantasy-land/map': f => Async(g => run(x => g(f(x))))
  };
}

Async['@@type'] = 'My/Async';
```

There's no prototype to be seen. :)

Furthermore, this approach makes `Async` a type representative, allowing `S.is(Async, Async.of(1))` to evaluate `true` (once `S.is` has been updated).

This change aligns with Fantasy Land's treatment of [type representatives][3]:

> If a type provides a type representative, each member of the type must have a `constructor` property which is a reference to the type representative.

:warning: If we decide to move `@@type` we'll need to increment the major version number.


[1]: https://gitter.im/sanctuary-js/sanctuary?at=584c3e8cc29531ac5d47629d
[2]: https://sanctuary.js.org/#is
[3]: https://github.com/fantasyland/fantasy-land#type-representatives

safareli:
👍 

if we move `@@type` I think other sanctuary types should also be update.

I think `@@type` should be defined on TypeRep as this information is same for all instances of the type and by sticking it in prototype makes it directly visible in all instances of the type.

in Free i'm storing '@@union' in typeRep which contains information about tagged constructors, and it seems more natural then sticking it in prototype
```
const description = {
  Pure: ['a'],
  Roll: ['x', 'y'],
}
const name = 'Seq'
const Seq = union(name, description)
// all information about type is in TypeRep 
Seq['@@type'] === name
Seq['@@union'] === description
```
(i'm storing `@@union` to have one [`cata` and `toString`](https://github.com/safareli/free/blob/78ca5488853291f2b68ca5bb981b4b9a134303a0/src/union.js) function for all such types)

davidchambers:
> if we move `@@type` I think other sanctuary types should also be update.

`S.Either` and `S.Maybe` do you mean, or are there others?

> i'm storing `@@union` to have one [`cata` and `toString`](https://github.com/safareli/free/blob/78ca5488853291f2b68ca5bb981b4b9a134303a0/src/union.js) function for all such types

Very cool!

safareli:
yes `S.Either` and `S.Maybe`

davidchambers:
I'll open a pull request to update `S.Either` and `S.Maybe` once this pull request has been merged.

davidchambers:
Could I get an LGTM, @Avaq or @safareli?

test/index.js
```diff
@@ -172,15 +172,15 @@ test('TypeClass', function() {
   //  Bar :: TypeClass
   var Bar = Z.TypeClass('my-package/Bar', [Foo], hasMethod('bar'));
 
-  eq(Foo['@@type'], 'sanctuary-type-classes/TypeClass');
+  eq(Foo.constructor['@@type'], 'sanctuary-type-classes/TypeClass');
```
safareli: I think `Z.type` could be used here, which is already tested and we could avoid direct access to ['@@type'] here.
```js
eq(Z.type(Foo), 'sanctuary-type-classes/TypeClass');
```

safareli: `sanctuary` depends on `sanctuary-def`
`sanctuary-def` depends on `sanctuary-type-classes`
so  if we export it here we can use it in sanctuary-def and sanctuary

davidchambers: Indeed. My only concern is that `@@type` has very little to do with Fantasy Land, so a `@@type`-specific function would be a little out of place here. I'm considering a separate package, sanctuary-type, which would just provide this function. I like to idea of a readme dedicated to explaining the problem `@@type` solves and the simple algorithm used to determine a value's type identifier.

safareli: it would be like a mini "spec"

davidchambers: > it would be like a mini "spec"

Indeed! See sanctuary-js/sanctuary-type-identifiers#1. :)

davidchambers: This pull request now uses the `type` function exported by [sanctuary-type-identifiers][1]. :)

:zap:


[1]: https://github.com/sanctuary-js/sanctuary-type-identifiers

PR #24, opened 2016-12-24
fantasy-land@2.2.0

davidchambers:
https://github.com/fantasyland/fantasy-land/compare/v2.1.0...v2.2.0

The significant pull request in the range above is fantasyland/fantasy-land#213.

GitHub does a really nice job of highlighting the changes to the ASCII diagram!

### DIFF
--- a/index.js
+++ b/index.js
@@ -31,30 +31,30 @@
 //. ## Type-class hierarchy
 //.
 //. <pre>
-//:  Setoid   Semigroup   Foldable        Functor         Extend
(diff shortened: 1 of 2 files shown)


svozza:
Interesting, a friend of mine and I had been looking at the FL spec a few weeks ago and were really confused by Functor's relationship to Comonad.

safareli:
btw technically it's braking change i think

davidchambers:
> btw technically it's braking change i think

You're right. Fortunately I realized this too. The change was released in v2.0.0. :)

PR #26, opened 2016-12-27
circumvent Safari bug

davidchambers:
In strict mode `function x(x) {}` is a syntax error in certain Safari versions (mishoo/UglifyJS2#179). Safari 10.0.2 does not have this bug, but the Sanctuary test suite runs against Safari 8 as well. Score another point for @Shard; real browser tests are proving very useful!

I'm targeting a temporary `v1` branch. After merging this pull request I'll release v1.3.1. I'll then merge `v1` into `master`, release v2.0.1, then delete `v1`.


safareli:
how adding `_` to argument names fix bug in safari?

PR #28, opened 2016-12-31
fantasy-land@3.0.0

davidchambers:
This is a breaking change, but a nice change I think. Do you agree?

### DIFF
--- a/test/List.js
+++ b/test/List.js
@@ -73,10 +73,10 @@ List.prototype[FL.reduce] = function(f, x) {
     Z.reduce(f, f(x, this.head), this.tail);
 };
 
-List.prototype[FL.traverse] = function(f, of) {
+List.prototype[FL.traverse] = function(typeRep, f) {
   return this.tag === 'Nil' ?
-    of(Nil) :
-    Z.ap(Z.map(curry2(Cons), f(this.head)), Z.traverse(of, f, this.tail));
+    Z.of(typeRep, Nil) :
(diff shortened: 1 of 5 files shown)


test/List.js
```diff
@@ -76,7 +76,7 @@ List.prototype[FL.reduce] = function(f, x) {
 List.prototype[FL.traverse] = function(f, of) {
   return this.tag === 'Nil' ?
     of(Nil) :
-    Z.ap(Z.map(curry2(Cons), f(this.head)), Z.traverse(of, f, this.tail));
+    Z.ap(Z.map(curry2(Cons), f(this.head)), this.tail[FL.traverse](f, of));
```
safareli: This part was a bit confusing, why not call `Z.traverse(List, f, this.tail)`?

davidchambers: `of` is not necessarily `List['fantasy-land/of']` (in fact, it almost certainly is not). We have no way of getting from `of` to a type representative, so we're forced to use the `of`-based API rather than our API based on type representatives.

If we update [`fantasy-land/traverse`][1] to take a type representative this problem will disappear. ;)


[1]: https://github.com/fantasyland/fantasy-land#traverse-method

davidchambers: fantasyland/fantasy-land#220

davidchambers: :zap:

safareli:
⭐️

PR #30, opened 2017-01-31
fix incorrect use of sequence in example

safareli:
sequence should take `TypeRep` and in example we are passing a function


davidchambers:
Thanks! I updated the similar `traverse` example in #28 but missed this one.

davidchambers:
I published v3.0.1 so the readme now reflects this fix. :)

PR #36, opened 2017-03-04
fantasy-land@3.1.0

davidchambers:
https://github.com/fantasyland/fantasy-land/compare/v3.0.0...v3.1.0

How does this look, @rjmk? None of JavaScript's built-in types can satisfy Contravariant, as far as I can see.

### DIFF
--- a/index.js
+++ b/index.js
@@ -31,8 +31,8 @@
 //. ## Type-class hierarchy
 //.
 //. <pre>
-//:  Setoid   Semigroup   Foldable        Functor
-//: (equals)   (concat)   (reduce)         (map)
+//:  Setoid   Semigroup   Foldable        Functor      Contravariant
+//: (equals)   (concat)   (reduce)         (map)        (contramap)
 //:               |           \         / | | | | \
 //:               |            \       /  | | | |  \
 //:               |             \     /   | | | |   \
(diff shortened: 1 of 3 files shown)


safareli:
function can be Contravariant in it's first argument:
```js
// -- for example lets look at functions from `a -> Boolean`
// Pred a = Pred { run:: (a -> Boolean)}

// it consumes Numbers 
// :: Pred Number
// :: Number ->  Boolean
const eq1 = a => 1 == 1

// :: ArrayLike a -> Number
const length = a => a.length

// contramap :: Contravariant f => f a ~> (b -> a) -> f b

// eq1.contramap :: Pred Number ~> (ArrayLike z -> Number) -> Pred (ArrayLike z)

// :: Pred ArrayLike z
// :: ArrayLike z -> Boolean
const lengthEq1 = eq1.contramap(length)


// -- implementation
//  Function$prototype$contramap :: (a -> b) ~> (c -> a) -> (c -> b)
function Function$prototype$contramap(f) {
  // :: a -> b
  var contravariant = this;
  // :: c -> b
  return function(x) { return contravariant(f(x)); };
}

//  Function$prototype$map :: (a -> b) ~> (b -> c) -> (a -> c)
function Function$prototype$map(f) {
  var functor = this;
  return function(x) { return f(functor(x)); };
}

// -- btw for function i think this is true:
// -- j.promap(f, g) = j.contramap(f).map(g)
//
//  Function$prototype$promap :: (b -> c) ~> (a -> b, c -> d) -> (a -> d)
function Function$prototype$promap(f, g) {
  var profunctor = this;
  return function(x) { return g(profunctor(f(x))); };
}
```

rjmk:
So in Fantasy Land we decided not to consider profunctors (like function) to be contravariant. Mostly because its too confusing when we use Haskell-style type notation. (You'd be saying something like `Contravariant (-> b)` which uses (invented) type operator sections).

I don't think there's a huge problem with Sanctuary taking an alternative route

davidchambers:
Thanks for the comment, Rafe. :)

PR #37, opened 2017-03-06
replace Monoid dependency with Alternative dependency for ‘filterM’

davidchambers:
:warning: This is a breaking change.

It became apparent in sanctuary-js/sanctuary#359 that `Z.filterM(odd, Just(4))` is invalid due to the `Monoid (m a)` constraint. `Just(4)` cannot satisfy this constraint as it cannot satisfy the requirements of Semigroup. `Just(4)` can, though, satisfy the requirements of Alternative.

This will allow `S.filterM` to operate on values of type `Maybe a` even with type checking enabled.

### DIFF
--- a/test/List.js
+++ b/test/List.js
@@ -35,6 +35,8 @@ List[FL.empty] = function() { return Nil; };
 
 List[FL.of] = function(x) { return Cons(x, Nil); };
 
+List[FL.zero] = List[FL.empty];
+
 List.prototype[FL.equals] = function(other) {
   return this.tag === 'Nil' ?
     other.tag === 'Nil' :
@@ -66,6 +68,8 @@ List.prototype[FL.chain] = function(f) {
     Nil :
     Z.concat(f(this.head), Z.chain(f, this.tail));
 };
+
+List.prototype[FL.alt] = List.prototype[FL.concat];
 
 List.prototype[FL.reduce] = function(f, x) {
   return this.tag === 'Nil' ?
--- a/index.js
+++ b/index.js
@@ -1333,8 +1333,8 @@
   //.
   //. Filters its second argument in accordance with the given predicate.
   //.
-  //. This function is derived from [`empty`](#empty), [`of`](#of), and
-  //. [`reduce`](#reduce).
+  //. This function is derived from [`concat`](#concat), [`empty`](#empty),
+  //. [`of`](#of), and [`reduce`](#reduce).
   //.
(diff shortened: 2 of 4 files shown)


test/List.js
```diff
@@ -67,6 +69,10 @@ List.prototype[FL.chain] = function(f) {
     Z.concat(f(this.head), Z.chain(f, this.tail));
 };
 
+List.prototype[FL.alt] = function(other) {
```
safareli: In [haskell](https://hackage.haskell.org/package/base-4.9.1.0/docs/src/GHC.Base.html#line-756) alt for list is concat:

```hs
instance Alternative [] where
    empty = []
    (<|>) = (++)
```

davidchambers: Good catch. I find this counter-intuitive, so I tend to forget. Thanks for the reminder. :)

safareli: Even though, both of the implementations will conform to Alternative laws

syaiful6:
We maybe need to relax the constraint, instead of Alternative, we can just take Plus constraint. List can implements Plus, but it can't implements Alternative, as it violate the distribute laws of Alternative as Fantasy Land defined. Because ```x.ap(f.alt(g))``` is not equivalent to ```x.ap(f).alt(x.ap(g))``` in List.

safareli:
`x.ap(f.alt(g))` is equivalent to `x.ap(f).alt(x.ap(g))`

for example:
```js
const x = [1]
const f = [a => a+1]
const g = [a => a*3]
f.alt(g) = [a => a+1, a => a*3]
x.ap(f.alt(g)) = [(a => a+1)(1), (a => a*3)(1)]
x.ap(f.alt(g)) = [2, 3]
x.ap(f) = [2]
x.ap(g) = [3]
x.ap(f).alt(x.ap(g)) = alt([2], [3]) = [2,3]

**Comment by syaiful6** [comment]
ah, correct..

**Comment by davidchambers** [comment]
I modified Irakli's example to convince myself that the equivalence is not limited to singleton lists:

```haskell
> ([(+ 1),(+ 9)] <|> [(* 3),(* 9)]) <*> [1,2,3]
[2,3,4,10,11,12,3,6,9,9,18,27]

> [(+ 1),(+ 9)] <*> [1,2,3] <|> [(* 3),(* 9)] <*> [1,2,3]
[2,3,4,10,11,12,3,6,9,9,18,27]
```

PR #38, opened 2017-03-09
getBoundMethod now checks whether or not a Constructor is a function

gabejohnson:
See https://github.com/sanctuary-js/sanctuary/issues/361

I added a `throws` function to check that the proper exception was being thrown. I also updated `test/Maybe` to implement `of` and `empty` as its constructor is not a function and is thus a good candidate for testing this code path.

### DIFF
--- a/index.js
+++ b/index.js
@@ -196,14 +196,25 @@
     return _funcPath(false, path, implementations);
   }
 
+  //  functionName :: Function -> String
+  var functionName = 'name' in function f() {} ?
+    function functionName(f) { return f.name; } :
+    /* istanbul ignore next */
+    function functionName(f) {
+      var match = /function (\w*)/.exec(f);
+      return match == null ? '' : match[1];
+    };
+
   //  $ :: (String, Array TypeClass, StrMap (Array Location)) -> TypeClass
   function $(_name, dependencies, requirements) {
     function getBoundMethod(_name) {
       var name = 'fantasy-land/' + _name;
       return requirements[_name] === Constructor ?
         function(typeRep) {
-          return funcPath([name], typeRep) ||
-                 implPath([/function (\w*)/.exec(typeRep)[1], name]);
+          var f = funcPath([name], typeRep);
+          return f == null && typeof typeRep === 'function' ?
+            implPath([functionName(typeRep), name]) :
+            f;
         } :
         function(x) {
           var isPrototype = x != null &&
--- a/test/index.js
+++ b/test/index.js
@@ -546,6 +546,7 @@ test('empty', function() {
   eq(Z.empty(Array), []);
   eq(Z.empty(Object), {});
   eq(Z.empty(List), Nil);
+  eq(Z.empty(Maybe), Nothing);
 });
 
 test('map', function() {
@@ -647,6 +648,7 @@ test('of', function() {
   eq(Z.of(Function, 42)(null), 42);
   eq(Z.of(Identity, 42), Identity(42));
   eq(Z.of(List, 42), Cons(42, Nil));
+  eq(Z.of(Maybe, 42), Just(42));
 });
 
 test('chain', function() {
--- a/test/Maybe.js
+++ b/test/Maybe.js
@@ -19,9 +19,9 @@ Maybe['@@type'] = 'sanctuary-type-classes/Maybe';
 
 Maybe.Nothing = new _Maybe('Nothing');
 
-Maybe.Just = function(x) { return new _Maybe('Just', x); };
+Maybe[FL.of] = Maybe.Just = function(x) { return new _Maybe('Just', x); };
 
-Maybe[FL.zero] = function() { return Maybe.Nothing; };
+Maybe[FL.empty] = Maybe[FL.zero] = function() { return Maybe.Nothing; };
 
 Maybe.prototype[FL.equals] = function(other) {
   return this.isNothing ? other.isNothing


index.js
```diff
@@ -203,7 +203,9 @@
       return requirements[_name] === Constructor ?
         function(typeRep) {
           return funcPath([name], typeRep) ||
-                 implPath([/function (\w*)/.exec(typeRep)[1], name]);
+                 implPath(typeof typeRep === 'function' ?
+                          [/function (\w*)/.exec(typeRep)[1], name] :
+                          [name]);
```
davidchambers: I don't think this is correct. We're expecting something like `implPath(['Array', 'fantasy-land/of'])` which is roughly equivalent to `implementations['Array']['fantasy-land/of']`. If the type representative is not a function it doesn't make sense to attempt something like `implementations['fantasy-land/of']`.

Also, the `typeof` check does not prevent `/function (\w*)/.exec(x => x)[1]` from throwing.

I think we want something like this:

```javascript
var f = funcPath([name], typeRep);
if (f == null && typeof typeRep === 'function') {
  var match = /function (\w*)/.exec(typeRep);
  if (match != null) f = implPath([match[1], name]);
}
return f;
```

gabejohnson: I'm also going to add a test for `Function#name` for environments that support it.

index.js
```diff
@@ -202,8 +202,20 @@
...
+          var f = funcPath([name], typeRep);
+          if (f == null && typeof typeRep === 'function') {
+            var fName;
+            /* istanbul ignore else */
+            if (typeRep.name) {
+              fName = typeRep.name;
+            } else {
+              var match = /function (\w*)/.exec(typeRep);
+              if (match != null) fName = match[1];
+            }
```
davidchambers: Let's move this to a separate function:

```javascript
//  functionName :: Function -> Nullable String
var functionName = 'name' in function f() {} ?
  function functionName(f) { return f.name; } :
  function functionName(f) {
    var match = /function (\w*)/.exec(f);
    return match && match[1];
  };
```

test/index.js
```diff
@@ -813,3 +816,9 @@ test('contramap', function() {
 
   eq(Z.contramap(length, inc)('abc'), 4);
 });
+
+test('errors', function() {
+  throws(function() { Z.chainRec(Maybe); },
+         TypeError,
+         'ChainRec.methods.chainRec(...) is not a function');
+});
```
davidchambers: I'm not sure it's a good idea to make assertions about errors thrown unintentionally. The error message may differ from one implementation to the next.

gabejohnson: Do you have any ideas for a better regression test?

davidchambers: I don't, I'm sorry.

I don't think a regression test is crucial, though, since we're replacing one "unexpected" exception with another.

gabejohnson: Shall I remove the `throws` implementation and other changes related to that test then?

davidchambers: That seems reasonable to me.

index.js
```diff
@@ -202,8 +202,18 @@
...
-                 implPath([/function (\w*)/.exec(typeRep)[1], name]);
+          var f = funcPath([name], typeRep);
+          if (f == null && typeof typeRep === 'function') {
+            /* istanbul ignore next */
+            var functionName = 'name' in function() {} ?
+                function functionName(f) { return f.name; } :
+                function functionName(f) {
+                  var match = /function (\w*)/.exec(f);
+                  return match && match[1];
+                };
```
davidchambers: Let's make this a top-level function. We needn't define it more than once, and moving it will make this part of the file clearer. :)

gabejohnson: Do we want the signature to be:

`Function -> Nullable String` or `Function -> String`?

I'm partial to the latter, but I know the code will cope with both.

davidchambers: Why do you prefer `Function -> String`, @gabejohnson? Are you assuming that all environments which support arrow functions also support `Function#name`?

gabejohnson: I believe that is the case (if you don't count es2015+ -> es5 transpilation). But it's more about preferring invariants. I suppose I could just do a check on the return value and only place it in the array if it's a string.

davidchambers: > I suppose I could just do a check on the return value and only place it in the array if it's a string.

If `functionName(typeRep)` returns `null` we should not apply `implPath` at all. ;)

safareli: if empty string and null/undefined are treated in a same way we can remove Nullable and will get only 1 value (empty string) for absence of function name instead of the 3.

index.js
```diff
@@ -196,14 +196,28 @@
     return _funcPath(false, path, implementations);
   }
 
+  /* istanbul ignore next */
+  // functionName :: Function -> Nullable String
```
davidchambers: Please add a second space after the `//`.

index.js
```diff
@@ -196,14 +196,28 @@
...
+  var functionName = 'name' in function() {} ?
```
davidchambers: Even though it's unnecessary, let's use `function f() {}` here.

index.js
```diff
@@ -196,14 +196,28 @@
...
   function $(_name, dependencies, requirements) {
     function getBoundMethod(_name) {
       var name = 'fantasy-land/' + _name;
       return requirements[_name] === Constructor ?
         function(typeRep) {
-          return funcPath([name], typeRep) ||
-                 implPath([/function (\w*)/.exec(typeRep)[1], name]);
+          var f = funcPath([name], typeRep);
+          if (f == null && typeof typeRep === 'function') {
+            var fn = functionName(typeRep);
```
davidchambers: `fn` makes me think "function". Let's name this `typeIdent` instead.

gabejohnson:
Should be ready to go now.

index.js
```diff
@@ -196,14 +196,28 @@
     return _funcPath(false, path, implementations);
   }
 
+  /* istanbul ignore next */
```
davidchambers: Let's move this to the "else" branch.

index.js
```diff
@@ -196,14 +196,28 @@
...
+  //  functionName :: Function -> Nullable String
+  var functionName = 'name' in function f() {} ?
+      function functionName(f) { return f.name; } :
+      function functionName(f) {
+        var match = /function (\w*)/.exec(f);
+        return match && match[1];
+      };
```
davidchambers: Please reduce the indentation of the fives lines above from four spaces to two spaces.

index.js
```diff
@@ -196,14 +196,28 @@
...
       var name = 'fantasy-land/' + _name;
       return requirements[_name] === Constructor ?
         function(typeRep) {
-          return funcPath([name], typeRep) ||
-                 implPath([/function (\w*)/.exec(typeRep)[1], name]);
+          var f = funcPath([name], typeRep);
+          if (f == null && typeof typeRep === 'function') {
+            var typeIdent = functionName(typeRep);
+            /* istanbul ignore else */
+            if (typeIdent) f = implPath([typeIdent, name]);
```
davidchambers: Let's use `typeIdent != null` as the condition.

The other option is to follow Irakli's suggestion by having `functionName` return `''` rather than `null` in the failure case. We could then remove the conditional logic above:

```javascript
f = implPath([functionName(typeRep), name]);
```

I like this idea now that we're using `Function#name` if available, since `(function() {}).name` and `(x => x).name` evaluate to `''` in modern environments.

gabejohnson: > The other option is to follow Irakli's suggestion by having functionName return '' rather than null in the failure case.

This is what I had in mind when I asked about the signature for `functionName`.

> Do we want the signature to be:
Function -> Nullable String or Function -> String?
I'm partial to the latter, but I know the code will cope with both.

The reason I would want to keep the check is because it will save us a loop in [`_funcPath`](https://github.com/gabejohnson/sanctuary-type-classes/blob/d2965b00f80d3c9dfb9d2da3906caa383ec6f2fb/index.js#L181)

davidchambers: I'd like to ensure that the condition evaluates to a Boolean value. What about the following?

```diff
-  //  functionName :: Function -> Nullable String
+  //  functionName :: Function -> String
```

```diff
-            if (typeIdent) f = implPath([typeIdent, name]);
+            if (typeIdent !== '') f = implPath([typeIdent, name]);
```

gabejohnson:
@davidchambers I've pushed up the changes.

index.js
```diff
@@ -196,14 +196,28 @@
     return _funcPath(false, path, implementations);
   }
 
+  //  functionName :: Function -> String
+  var functionName = 'name' in function f() {} ?
+    function functionName(f) { return f.name; } :
+    /* istanbul ignore next */
+    function functionName(f) {
+      var match = /function (\w*)/.exec(f);
+      return (match || '') && match[1];
```
davidchambers: I think `match == null ? '' : match[1]` would be clearer. Currently both `||` and `&&` are "applied" in the failure case.

gabejohnson: Updated

index.js
```diff
@@ -196,14 +196,28 @@
...
       var name = 'fantasy-land/' + _name;
       return requirements[_name] === Constructor ?
         function(typeRep) {
-          return funcPath([name], typeRep) ||
-                 implPath([/function (\w*)/.exec(typeRep)[1], name]);
+          var f = funcPath([name], typeRep);
+          if (f == null && typeof typeRep === 'function') {
+            var typeIdent = functionName(typeRep);
+            /* istanbul ignore else */
+            if (typeIdent !== '') f = implPath([typeIdent, name]);
```
davidchambers: I'd like to test the else branch rather than ignore it. I haven't been able to come up with a test case, though, due to the extreme lengths that are taken to determine `Function#name`. For example:

```javascript
> ({constructor: function() {}}).constructor.name
'constructor'
```

If there isn't a scenario in which `typeIdent` will be `''` in an environment which provides `Function#name`, the guard isn't helpful. We could write this instead:

```javascript
var f = funcPath([name], typeRep);
return f == null && typeof typeRep === 'function' ?
  implPath([functionName(typeRep), name]) :
  f;
```

Avaq: `({constructor: (function(){ return function(){} })()}).constructor.name`?

gabejohnson: Not all environments that support `Function#name` do so equally. For instance, Chrome supports
```js
> ({constructor: function() {}}).constructor.name
'constructor'
```
but FireFox 52 does not.

davidchambers: That's clever, Aldwin. Such code is unlikely to exist in practice, so optimizing the code path for `''` isn't advantageous.

davidchambers: Thanks for the information about Firefox, @gabejohnson. I'd still prefer to avoid a micro-optimization which will lead to an unreachable code branch in the foreseeable future.

gabejohnson: Updated

Issue #43, opened 2017-04-08
fantasy-land@3.2.0


davidchambers:
https://github.com/fantasyland/fantasy-land/compare/v3.1.0...v3.2.0

I believe that @gabejohnson is already working on this change. Could you confirm this, Gabe?

gabejohnson:
That is correct. I have already implemented `Ord` for `Null`, `Undefined`, `Boolean`, `Number`, `Date`, and `String`. I plan to add `Array` and `Arguments` this weekend (they are assumed to have the same semantics).

safareli:
I'm a bit lost, what's difference between #46 and #45

davidchambers:
> I'm a bit lost, what's difference between #46 and #45

These pull requests were submitted independently, and the patches are largely duplicative (although #45 contains additions unrelated to fantasy-land@3.2.0 which could become separate pull requests).

I decided to keep both pull requests open as there were significant differences between them initially. They are gradually converging. At some point we'll close one and put our combined energy behind the other.

PR #45, opened 2017-04-09
Add missing instances (mostly `Ord`)

i-am-tom:
Several missing instances have been added. The majority are for `Ord`, though there is also an `Alt` and `Semigroup` instance for `Function`. Basically a port of https://github.com/fantasyland/fantasy-land/pull/244

I've hopefully done a satisfactory job of tests, format, etc!

### DIFF
--- a/index.js
+++ b/index.js
@@ -518,7 +518,9 @@
 
   //  Boolean$prototype$equals :: Boolean ~> Boolean -> Boolean
   function Boolean$prototype$equals(other) {
-    return typeof other === typeof this && other.valueOf() === this.valueOf();
+    return typeof this === 'object' ?
+      equals(this.valueOf(), other.valueOf()) :
+      this === other;
   }
 
   //  Number$prototype$toString :: Number ~> () -> String
@@ -530,11 +532,9 @@
 
   //  Number$prototype$equals :: Number ~> Number -> Boolean
   function Number$prototype$equals(other) {
-    return typeof other === 'object' ?
-      typeof this === 'object' &&
-        equals(this.valueOf(), other.valueOf()) :
-      isNaN(other) && isNaN(this) ||
-        other === this && 1 / other === 1 / this;
+    return typeof this === 'object' ?
+      equals(this.valueOf(), other.valueOf()) :
+      isNaN(this) && isNaN(other) || this === other && 1 / this === 1 / other;
   }
 
   //  Date$prototype$toString :: Date ~> () -> String
@@ -580,7 +580,9 @@
 
   //  String$prototype$equals :: String ~> String -> Boolean
   function String$prototype$equals(other) {
-    return typeof other === typeof this && other.valueOf() === this.valueOf();
+    return typeof this === 'object' ?
+      equals(this.valueOf(), other.valueOf()) :
+      this === other;
   }
 
   //  String$prototype$concat :: String ~> String -> String
@@ -1016,9 +1018,7 @@
     var $pairs = [];
 
     return function equals(x, y) {
-      if (type(x) !== type(y)) {
-        return false;
-      }
+      if (typeof x !== typeof y || type(x) !== type(y)) return false;
 
       //  This algorithm for comparing circular data structures was
       //  suggested in <http://stackoverflow.com/a/40622794/312785>.
--- a/index.js
+++ b/index.js
@@ -255,6 +255,16 @@
   //. ```
   var Setoid = $('Setoid', [], {equals: Value});
 
+  //# Ord :: TypeClass
+  //.
+  //. `TypeClass` value for [Ord][].
+  //.
+  //. ```javascript
+  //. > Ord.test(null)
+  //. true
+  //. ```
+  var Ord = $('Ord', [Setoid], {lte: Value});
+
   //# Semigroup :: TypeClass
   //.
   //. `TypeClass` value for [Semigroup][].
@@ -499,6 +509,11 @@
     return true;
   }
 
+  //  Null$prototype$lte :: Null ~> Null -> Boolean
+  function Null$prototype$lte(other) {
+    return true;
+  }
+
   //  Undefined$prototype$toString :: Undefined ~> () -> String
   function Undefined$prototype$toString() {
     return 'undefined';
@@ -506,6 +521,11 @@
 
   //  Undefined$prototype$equals :: Undefined ~> Undefined -> Boolean
   function Undefined$prototype$equals(other) {
+    return true;
+  }
+
+  //  Undefined$prototype$lte :: Undefined ~> Undefined -> Boolean
+  function Undefined$prototype$lte(other) {
     return true;
   }
 
@@ -523,6 +543,13 @@
       this === other;
   }
 
+  //  Boolean$prototype$lte :: Boolean ~> Boolean -> Boolean
+  function Boolean$prototype$lte(other) {
+    return typeof this === 'object' ?
+      lte(this.valueOf(), other.valueOf()) :
+      this <= other;
+  }
+
   //  Number$prototype$toString :: Number ~> () -> String
   function Number$prototype$toString() {
     return typeof this === 'object' ?
@@ -534,7 +561,14 @@
   function Number$prototype$equals(other) {
     return typeof this === 'object' ?
       equals(this.valueOf(), other.valueOf()) :
-      isNaN(this) && isNaN(other) || this === other && 1 / this === 1 / other;
+      isNaN(this) && isNaN(other) || this === other;
+  }
+
+  //  Number$prototype$lte :: Number ~> Number -> Boolean
+  function Number$prototype$lte(other) {
+    return typeof this === 'object' ?
+      lte(this.valueOf(), other.valueOf()) :
+      isNaN(other) && isNaN(this) || this <= other;
   }
 
   //  Date$prototype$toString :: Date ~> () -> String
@@ -546,6 +580,11 @@
   //  Date$prototype$equals :: Date ~> Date -> Boolean
   function Date$prototype$equals(other) {
     return equals(this.valueOf(), other.valueOf());
+  }
+
+  //  Date$prototype$lte :: Date ~> Date -> Boolean
+  function Date$prototype$lte(other) {
+    return lte(this.valueOf(), other.valueOf());
   }
 
   //  RegExp$prototype$equals :: RegExp ~> RegExp -> Boolean
@@ -583,6 +622,13 @@
     return typeof this === 'object' ?
       equals(this.valueOf(), other.valueOf()) :
       this === other;
+  }
+
+  //  String$prototype$lte :: String ~> String -> Boolean
+  function String$prototype$lte(other) {
+    return typeof this === 'object' ?
+      lte(this.valueOf(), other.valueOf()) :
+      this <= other;
   }
 
   //  String$prototype$concat :: String ~> String -> String
@@ -639,6 +685,19 @@
     for (var idx = 0; idx < this.length; idx += 1) {
       if (!equals(this[idx], other[idx])) return false;
     }
+    return true;
+  }
+
+  //  Array$prototype$lte :: Array a ~> Array a -> Boolean
+  function Array$prototype$lte(other) {
+    for (var idx = 0; idx < this.length; idx += 1) {
+      if (idx >= other.length) return false;
+
+      if (!equals(this[idx], other[idx])) {
+        return lte(this[idx], other[idx]);
+      }
+    }
+
     return true;
   }
 
@@ -709,6 +768,11 @@
   //  Arguments$prototype$equals :: Arguments ~> Arguments -> Boolean
   function Arguments$prototype$equals(other) {
     return Array$prototype$equals.call(this, other);
+  }
+
+  //  Arguments$prototype$lte :: Arguments ~> Arguments -> Boolean
+  function Arguments$prototype$lte(other) {
+    return Array$prototype$lte.call(this, other);
   }
 
   //  Error$prototype$toString :: Error ~> () -> String
@@ -828,6 +892,18 @@
     return function(x) { return f(chain(x))(x); };
   }
 
+  //  Function$prototype$alt :: Alt a => (x -> a) ~> (x -> a) -> x -> a
+  function Function$prototype$alt(f) {
+    var alt_ = this;
+    return function(x) { return alt(alt_(x), f(x)); };
+  }
+
+  //  Function$prototype$concat :: Semigroup a => (x -> a) ~> (x -> a) -> x -> a
+  function Function$prototype$concat(f) {
+    var semigroup = this;
+    return function(x) { return concat(semigroup(x), f(x)); };
+  }
+
   //  Function$prototype$contramap :: (b -> c) ~> (a -> b) -> (a -> c)
   function Function$prototype$contramap(f) {
     var contravariant = this;
@@ -839,31 +915,36 @@
     Null: {
       prototype: {
         toString:                   Null$prototype$toString,
-        'fantasy-land/equals':      Null$prototype$equals
+        'fantasy-land/equals':      Null$prototype$equals,
+        'fantasy-land/lte':         Null$prototype$lte
       }
     },
     Undefined: {
       prototype: {
         toString:                   Undefined$prototype$toString,
-        'fantasy-land/equals':      Undefined$prototype$equals
+        'fantasy-land/equals':      Undefined$prototype$equals,
+        'fantasy-land/lte':         Undefined$prototype$lte
       }
     },
     Boolean: {
       prototype: {
         toString:                   Boolean$prototype$toString,
-        'fantasy-land/equals':      Boolean$prototype$equals
+        'fantasy-land/equals':      Boolean$prototype$equals,
+        'fantasy-land/lte':         Boolean$prototype$lte
       }
     },
     Number: {
       prototype: {
         toString:                   Number$prototype$toString,
-        'fantasy-land/equals':      Number$prototype$equals
+        'fantasy-land/equals':      Number$prototype$equals,
+        'fantasy-land/lte':         Number$prototype$lte
       }
     },
     Date: {
       prototype: {
         toString:                   Date$prototype$toString,
-        'fantasy-land/equals':      Date$prototype$equals
+        'fantasy-land/equals':      Date$prototype$equals,
+        'fantasy-land/lte':         Date$prototype$lte
       }
     },
     RegExp: {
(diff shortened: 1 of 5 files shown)


i-am-tom:
Also, by the looks of it, the README is generated; I presume I therefore don't need to update it?

davidchambers:
Thanks for the pull request, @i-am-tom!

> Also, by the looks of it, the README is generated; I presume I therefore don't need to update it?

That's right. It will be generated when we run `make release-minor` or similar.

There's some duplication between this pull request and #46, but there are noteworthy differences between your implementations of certain methods and those Gabe wrote. We can discuss these and hopefully converge. :)

i-am-tom:
Sweet! Got there in the end :) Thanks, @davidchambers

index.js
```diff
@@ -521,6 +541,12 @@
     return typeof other === typeof this && other.valueOf() === this.valueOf();
   }
 
+  //  Boolean$prototype$lte :: Boolean ~> Boolean -> Boolean
+  function Boolean$prototype$lte(other) {
+    return typeof other === typeof this &&
+      this.valueOf() ? other.valueOf() : true;
+  }
+
```
davidchambers: This violates antisymmetry given our definition of `equals`:

```javascript
> Z.lte(false, new Boolean(false))
true

> Z.lte(new Boolean(false), false)
true

> Z.equals(false, new Boolean(false))
false
```

index.js
```diff
@@ -826,6 +888,18 @@
     return function(x) { return f(chain(x))(x); };
   }
 
+  //  Function$prototype$alt :: (x -> a) ~> (x -> a) -> x -> a
```
davidchambers: Should this signature include `Alt a =>`?

index.js
```diff
@@ -826,6 +888,18 @@
...
+  function Function$prototype$alt(f) {
+    var alt_ = this;
+    return function(x) { return alt(alt_(x), f(x)); };
+  }
+
+  //  Function$prototype$concat :: (x -> a) ~> (x -> a) -> x -> a
```
davidchambers: Should this signature include `Semigroup a =>`?

index.js
i-am-tom: Oo; will update accordingly.

index.js
i-am-tom: Yes, you're quite right!

test/List.js
```diff
@@ -45,6 +45,16 @@ List.prototype[FL.equals] = function(other) {
...
 };
 
+List.prototype[FL.lte] = function(other) {
+  if (this.tag === 'Nil') {
+    return true;
+  }
+
+  return Z.equals(other.head, this.head)
+    ? Z.lte(this.tail, other.tail)
+    : Z.lte(this.head, other.tail);
```
davidchambers: It doesn't make sense to compare a head with a tail, does it?

index.js
i-am-tom: Yep!

test/List.js
i-am-tom: >_< Typo

index.js
```diff
@@ -826,6 +888,18 @@
...
+  function Function$prototype$concat(f) {
+    var semigroup = this;
+    return function(x) { return concat(semigroup(x), f(x)); };
+  }
+
```
davidchambers: Please add tests for `Function$prototype$alt` and `Function$prototype$concat`.

In both cases I'm interested to know whether the order of arguments (to `alt` and `concat` respectively) is forced upon us by the laws, or whether both orders are lawful.

i-am-tom: No, only `this.concat(that)` - in both cases, we have associativity, but not necessarily commutativity. e.g.:

```haskell
Just [2] <> Just [3] == Just [2, 3]
(const $ Just [2]) <> (const $ Just [3]) == const $ Just [2, 3]
```

Just as we can't change ordering of `Array` concat within `Maybe` concat, we can't change ordering of `Array` concat within `Maybe` concat within `Function` concat!

davidchambers: Thanks for the explanation and example. It's nice to have decisions made for us. :)

i-am-tom: Hah, agreed! I've committed the other comments' changes, as well as some tests for alt/concat :)

index.js
```diff
@@ -521,6 +541,12 @@
     return typeof other === typeof this && other.valueOf() === this.valueOf();
   }
 
+  //  Boolean$prototype$lte :: Boolean ~> Boolean -> Boolean
+  function Boolean$prototype$lte(other) {
+    return typeof other === typeof this &&
+      (this.valueOf() ? other.valueOf() : true);
```
davidchambers: I believe `x ? y : true` can be written `!x || y`.

index.js
```diff
@@ -583,6 +621,12 @@
     return typeof other === typeof this && other.valueOf() === this.valueOf();
   }
 
+  //  String$prototype$lte :: String ~> String -> Boolean
+  function String$prototype$lte(other) {
+    return (typeof this === 'object' ? this.valueOf() : this)
+      <= (typeof other === 'object' ? other.valueOf() : other);
+  }
+
```
davidchambers: This violates antisymmetry:

```javascript
> Z.lte('abc', new String('abc'))
true

> Z.lte(new String('abc'), 'abc')
true

> Z.equals('abc', new String('abc'))
false
```

index.js
i-am-tom: It can; would you say that's more readable?

davidchambers: I find it easier to comprehend. It will evaluate `true` if `x` is `false` (regardless of `y`) or if `y` is `true` (regardless of `x`). Somehow this is closer to my mental model than the equivalent ternary expression.

index.js
```diff
@@ -521,6 +541,12 @@
     return typeof other === typeof this && other.valueOf() === this.valueOf();
   }
 
+  //  Boolean$prototype$lte :: Boolean ~> Boolean -> Boolean
+  function Boolean$prototype$lte(other) {
+    return typeof other === typeof this &&
+      (!this.valueOf() || other.valueOf());
+  }
+
```
davidchambers: This now violates totality:

```javascript
> Z.lte(false, new Boolean(false))
false

> Z.lte(new Boolean(false), false)
false
```

We should test the laws. ;)

i-am-tom: I thought this was expected behaviour? Differing constructors shouldn't be comparable?

davidchambers: This is an interesting conundrum. Let's discuss it in https://github.com/sanctuary-js/sanctuary-type-classes/pull/46#discussion_r110620730.

gabejohnson: I made the in https://github.com/sanctuary-js/sanctuary-type-classes/pull/46#discussion_r110638087

i-am-tom: Ok. Seems arbitrary to care about the difference between Boolean(true) and true - the Boolean algebra we're attempting to express wasn't written with four constructors in mind!

test/index.js
```diff
@@ -522,6 +537,86 @@ test('equals', function() {
...
+  eq(Z.lte(-0, -0), true);
+  eq(Z.lte(NaN, NaN), true);
+  eq(Z.lte(Infinity, Infinity), true);
+  eq(Z.lte(Infinity, -Infinity), false);
+  eq(Z.lte(-Infinity, Infinity), true);
+  eq(Z.lte(-Infinity, -Infinity), true);
+  eq(Z.lte(NaN, Math.PI), false);
+  eq(Z.lte(Math.PI, NaN), false);
+  eq(Z.lte(new Number(0), new Number(0)), true);
+  eq(Z.lte(new Number(0), new Number(-0)), true);
```
gabejohnson: This test is wrong. `lte(0, -0) === false` if `equals(0, -0) === false`.

test/index.js
```diff
@@ -522,6 +537,86 @@ test('equals', function() {
...
+  eq(Z.lte(new Boolean(false), new Boolean(false)), true);
+  eq(Z.lte(new Boolean(false), new Boolean(true)), true);
+  eq(Z.lte(new Boolean(true), new Boolean(false)), false);
+  eq(Z.lte(new Boolean(true), new Boolean(true)), true);
+  eq(Z.lte(false, new Boolean(false)), false);
+  eq(Z.lte(new Boolean(false), false), false);
+  eq(Z.lte(true, new Boolean(true)), false);
+  eq(Z.lte(new Boolean(true), true), false);
+  eq(Z.lte(0, 0), true);
+  eq(Z.lte(0, -0), true);
```
gabejohnson: See below

test/index.js
```diff
@@ -522,6 +537,86 @@ test('equals', function() {
...
+  eq(Z.lte(new Number(-0), new Number(-0)), true);
+  eq(Z.lte(new Number(NaN), new Number(NaN)), true);
+  eq(Z.lte(new Number(Infinity), new Number(Infinity)), true);
+  eq(Z.lte(new Number(Infinity), new Number(-Infinity)), false);
+  eq(Z.lte(new Number(-Infinity), new Number(Infinity)), true);
+  eq(Z.lte(new Number(-Infinity), new Number(-Infinity)), true);
+  eq(Z.lte(new Number(NaN), new Number(Math.PI)), false);
+  eq(Z.lte(new Number(Math.PI), new Number(NaN)), false);
+  eq(Z.lte(42, new Number(42)), true);
+  eq(Z.lte(new Number(42), 42), true);
```
gabejohnson: Same argument as above.

test/index.js
```diff
@@ -522,6 +537,86 @@ test('equals', function() {
...
+  eq(Z.lte(new String('abc'), new String('xyz')), true);
+  eq(Z.lte(new String('xyz'), new String('abc')), false);
+  eq(Z.lte('abc', new String('abc')), false);
+  eq(Z.lte(new String('abc'), 'abc'), false);
+  eq(Z.lte([], []), true);
+  eq(Z.lte([1, 2], [1, 2]), true);
+  eq(Z.lte([1, 2, 3], [1, 2]), false);
+  eq(Z.lte([1, 2], [1, 2, 3]), true);
+  eq(Z.lte([1, 2], [2, 1]), true);
+  eq(Z.lte([0], [-0]), true);
```
gabejohnson: See above

test/index.js
i-am-tom: In the set of Int (or Num, for that matter), there is no _value_ difference between `0` and `-0`. Just as with `true` vs. `Boolean(true)`, the difference is with JavaScript's incongruent way of expressing these values. Having `0` and `-0` as different values essentially means that, whichever group we're using, we'll end up with two identities. This doesn't seem like a desirable behaviour (although I might well be missing some obvious use cases!) - what do people think?

davidchambers: Preserving referential transparency seems more important to me than having a single identity. If `x` is equal to `y`, I don't want `1 / x` to give `Infinity` while `1 / y` gives `-Infinity`. One could argue, though, that division be zero is nonsensical. If you'd like to take that position, @i-am-tom, I'm open to considering the possibility of treating `0` and `-0` as equal.

index.js
safareli: `Function$prototype$alt :: (x -> a) ~> (x -> a) -> x -> a` here `a` is of kinde `*` but Alt is of kind `* -> *` so i think it should be `(Alt f) => (x -> f a) ~> (x -> f a) -> x -> f a` but not sure.

test/index.js
i-am-tom: @davidchambers: it would certainly make sense given the language we're in:

```javascript
> 0 === -0
true
```

Either decision seems to be a case of trading one weird exception for another, but even [PureScript's `Semiring` docs](https://pursuit.purescript.org/packages/purescript-prelude/3.0.0/docs/Data.Semiring) come with a proviso that `Infinity` is undefined behaviour. _They make an excellent point about overflow, too, which **certainly** violates referential transparency_. In particular, with boolean values, it's a shame to sacrifice the purity of a two-value algebra because of a language quirk that (to my surprise!) don't even matter to the _language_:

```javascript
> true === Boolean(true)
true
```

... especially with all the other inconsistencies. Don't get me started on float arithmetic 😆

davidchambers: `Boolean :: Any -> Boolean`. I think you meant to do this:

```javascript
> true === new Boolean(true)
false

> new Boolean(true) === new Boolean(true)
false
```

Following PureScript's lead sounds good to me. Please update the definition of `Number$prototype$equals` to consider `-0` equal to `0`.

i-am-tom: Ah, I see - false example, but the original point certainly still holds. I'll update the `Number` prototype now! 👍

gabejohnson:
@i-am-tom, you'll have to update the docs as well. Doc tests are run on node v6. Hence the build failure.

i-am-tom:
Of course, sorry. Went back to work and promptly forgot about it. Apologies - will fix now!

test/index.js
```diff
@@ -522,6 +533,86 @@ test('equals', function() {
...
+  eq(Z.lte(args(1, 2, 3), args(1, 2)), false);
+  eq(Z.lte(args(1, 2), args(1, 2, 3)), true);
+  eq(Z.lte(args(1, 2), args(2, 1)), true);
+  eq(Z.lte(args(0), args(-0)), true);
+  eq(Z.lte(args(NaN), args(NaN)), true);
+  eq(Z.lte(Identity(Identity(Identity(0))), Identity(Identity(Identity(0)))), true);
+  eq(Z.lte(Identity(Identity(Identity(0))), Identity(Identity(Identity(1)))), true);
+  eq(Z.lte(Identity(Identity(Identity(1))), Identity(Identity(Identity(0)))), false);
+  eq(Z.lte(Identity(Identity(Identity(1))), Identity(Identity(Identity(1)))), true);
+});
```
gabejohnson: @i-am-tom You should probably add some tests covering nested Arrays.

i-am-tom: @gabejohnson Can do! What's the motivation for doing so? If testing single-level arrays shows that `lte` is delegated to the inner type, what situation could that test catch? Curious, is all - I'll add the test later tonight.

gabejohnson: Sorry, I meant circular Arrays. If an Array contains a reference to itself you'll get stuck in an infinite loop. See the `equals` tests for examples.

i-am-tom: Ah, yes. I explicitly removed the mention of this from the `lte` docs. As there is, to my knowledge, no _pure_ way to create such a structure, I'm wasn't sure why it was being considered. However, if there's a reason I'm missing, I'm happy to update the code to deal with such a circumstance!

gabejohnson: Given your implementation of `Array$prototype$lte`, it may not be needed. I'm having trouble coming up with a test case that wouldn't be covered by the handling of circular structures in `equals`.

davidchambers:
@i-am-tom, please rebase this branch now that #47 has been merged.

i-am-tom:
Sorted; CI-ing now!

index.js
```diff
@@ -532,9 +558,17 @@
...
   //  Number$prototype$equals :: Number ~> Number -> Boolean
   function Number$prototype$equals(other) {
-    return typeof this === 'object' ?
-      equals(this.valueOf(), other.valueOf()) :
-      isNaN(this) && isNaN(other) || this === other && 1 / this === 1 / other;
+    return typeof other === 'object' ?
+      typeof this === 'object' &&
+        equals(this.valueOf(), other.valueOf()) :
+      isNaN(other) && isNaN(this) || this === other;
+  }
```
davidchambers: The only change to this function should be the removal of the `1 / this === 1 / other` check.

index.js
```diff
@@ -532,9 +558,17 @@
...
+
+  //  Number$prototype$lte :: Number ~> Number -> Boolean
+  function Number$prototype$lte(other) {
+    return isNaN(this) ? isNaN(other) :
+      (typeof this === 'object' ? this.valueOf() : this)
+        <= (typeof other === 'object' ? other.valueOf() : other);
   }
```
davidchambers: Let's use the simpler implementation that #47 enabled:

```javascript
//  Number$prototype$lte :: Number ~> Number -> Boolean
function Number$prototype$lte(other) {
  return typeof this === 'object' ?
    lte(this.valueOf(), other.valueOf()) :
    this <= other;
```

`Boolean$prototype$lte` and `String$prototype$lte` should be updated as well. :)

i-am-tom: Such a change breaks the following tests:

```
eq(Z.lte(NaN, NaN), true);
eq(Z.lte(new Number(NaN), new Number(NaN)), true);
eq(Z.lte(new Date(NaN), new Date(NaN)), true);
```

davidchambers: I missed something. These changes would need to be accompanied by a change to `lte` identical to the change to `equals` in #47.

i-am-tom: Among others, this now breaks both these tests:

```
eq(Z.lte(42, new Number(42)), true);
eq(Z.lte(new Number(42), 42), true);
```

Are we _sure_ it's worthwhile to differentiate here? It does go back to the point that we're making a boolean type with 4 different constructors, and now 2 * ∞(ish) numbers. If the above tests _break_, and this distinction is maintained, totality is broken.

Alternatively, we then have to pick whether `x <= new Number(x)` or `x >= new Number(x)`, and the same for other types such as (but not limited to) `Boolean`. That behaviour would be pretty unintuitive, because I then have a `false` value that may not be `<=` another `false`, depending on where I get those values.

It seems as though we either have to treat both forms of the same value as the same value, or not support constructor-borne primitives. What do you think?

davidchambers: The approach I suggested in https://github.com/sanctuary-js/sanctuary-type-classes/pull/46#discussion_r110643437 still seems best to me: provide definitions of `equals` and `lte` for both `NumberPrimitive` and `NumberObject`. This avoids arbitrarily deciding whether `0` comes before or after `new Number(0)`, because these values are incomparable.

i-am-tom: Ok, and hence delete the tests that try to compare them?

davidchambers: I wouldn't mind keeping the tests, to make our intentions clear to those reading the test suite in the future. The expected result of `Z.lte(42, new Number(42))` will need to be updated from `true` to `false`, of course.

i-am-tom: Cool, will do. Are we therefore happy with this behaviour: https://github.com/sanctuary-js/sanctuary-type-classes/pull/45#discussion_r110637365 ?

davidchambers: I believe the three of us now agree that any comparison between a `BooleanPrimitive` and a `BooleanObject` should evaluate `false`. :)

i-am-tom: Great! I'll try to get it done this afternoon. Thanks for sticking with this :)

index.js
davidchambers: @i-am-tom, would you mind opening two more pull requests? One for `Function#fantasy-land/alt` and one for `Function#fantasy-land/concat`. It would be nice to review and merge these changes separately as they're unrelated to #43.

davidchambers:
Let's close this pull request and focus our efforts on reviewing, polishing, and merging #46.

I'd love to see pull requests for `Function#fantasy-land/alt` and `Function#fantasy-land/concat`, @i-am-tom. :)

PR #46, opened 2017-04-10
fantasy-land@3.2.0

### LINKED ISSUE #43: fantasy-land@3.2.0

gabejohnson:
Closes #43

Implemented for the following:

* `null`
* `undefined`
* `Boolean`
* `Number`
* `Date`
* `String`
* `Array`
* `arguments`

I'm thinking about `Object` as well.

I still have many tests to add. The current tests are, for the most part, simply copied from the `equals` tests.

I'm having trouble with the implementation for `Array`. It could potentially be very expensive. Though I'm probably overlooking something simple.

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,13 +33,13 @@
 //. <pre>
 //:  Setoid   Semigroup   Foldable        Functor      Contravariant
 //: (equals)   (concat)   (reduce)         (map)        (contramap)
-//:               |           \         / | | | | \
-//:               |            \       /  | | | |  \
-//:               |             \     /   | | | |   \
-//:               |              \   /    | | | |    \
-//:               |               \ /     | | | |     \
-//:            Monoid         Traversable | | | |      \
-//:            (empty)        (traverse)  / | | \       \
+//:     |         |           \         / | | | | \
+//:     |         |            \       /  | | | |  \
+//:     |         |             \     /   | | | |   \
+//:     |         |              \   /    | | | |    \
+//:     |         |               \ /     | | | |     \
+//:    Ord     Monoid         Traversable | | | |      \
+//:   (lte)    (empty)        (traverse)  / | | \       \
 //:                                      /  | |  \       \
 //:                                     /   / \   \       \
 //:                             Profunctor /   \ Bifunctor \
@@ -110,6 +110,11 @@
     return function(y) {
       return [x, y];
     };
+  }
+
+  //  sameType :: (a, b) -> Boolean
+  function sameType(x, y) {
+    return typeof x === typeof y && type(x) === type(y);
   }
 
   //  type Iteration a = { value :: a, done :: Boolean }
@@ -254,6 +259,19 @@
   //. true
   //. ```
   var Setoid = $('Setoid', [], {equals: Value});
+
+  //# Ord :: TypeClass
+  //.
+  //. `TypeClass` value for [Ord][].
+  //.
+  //. ```javascript
+  //. > Ord.test(0)
+  //. true
+  //.
+  //. > Ord.test(Math.sqrt)
+  //. false
+  //. ```
+  var Ord = $('Ord', [Setoid], {lte: Value});
 
   //# Semigroup :: TypeClass
   //.
@@ -499,6 +517,11 @@
     return true;
   }
 
+  //  Null$prototype$lte :: Null ~> Null -> Boolean
+  function Null$prototype$lte(other) {
+    return true;
+  }
+
   //  Undefined$prototype$toString :: Undefined ~> () -> String
   function Undefined$prototype$toString() {
     return 'undefined';
@@ -506,6 +529,11 @@
 
   //  Undefined$prototype$equals :: Undefined ~> Undefined -> Boolean
   function Undefined$prototype$equals(other) {
+    return true;
+  }
+
+  //  Undefined$prototype$lte :: Undefined ~> Undefined -> Boolean
+  function Undefined$prototype$lte(other) {
     return true;
   }
 
@@ -523,6 +551,13 @@
       this === other;
   }
 
+  //  Boolean$prototype$lte :: Boolean ~> Boolean -> Boolean
+  function Boolean$prototype$lte(other) {
+    return typeof this === 'object' ?
+      lte(this.valueOf(), other.valueOf()) :
+      this === false || other === true;
+  }
+
   //  Number$prototype$toString :: Number ~> () -> String
   function Number$prototype$toString() {
     return typeof this === 'object' ?
@@ -534,7 +569,14 @@
   function Number$prototype$equals(other) {
     return typeof this === 'object' ?
       equals(this.valueOf(), other.valueOf()) :
-      isNaN(this) && isNaN(other) || this === other && 1 / this === 1 / other;
+      isNaN(this) && isNaN(other) || this === other;
+  }
+
+  //  Number$prototype$lte :: Number ~> Number -> Boolean
+  function Number$prototype$lte(other) {
+    return typeof this === 'object' ?
+      lte(this.valueOf(), other.valueOf()) :
+      isNaN(this) && isNaN(other) || this <= other;
   }
 
   //  Date$prototype$toString :: Date ~> () -> String
@@ -546,6 +588,11 @@
   //  Date$prototype$equals :: Date ~> Date -> Boolean
   function Date$prototype$equals(other) {
     return equals(this.valueOf(), other.valueOf());
+  }
+
+  //  Date$prototype$lte :: Date ~> Date -> Boolean
+  function Date$prototype$lte(other) {
+    return lte(this.valueOf(), other.valueOf());
   }
 
   //  RegExp$prototype$equals :: RegExp ~> RegExp -> Boolean
@@ -583,6 +630,13 @@
     return typeof this === 'object' ?
       equals(this.valueOf(), other.valueOf()) :
       this === other;
+  }
+
+  //  String$prototype$lte :: String ~> String -> Boolean
+  function String$prototype$lte(other) {
+    return typeof this === 'object' ?
+      lte(this.valueOf(), other.valueOf()) :
+      this <= other;
   }
 
   //  String$prototype$concat :: String ~> String -> String
@@ -640,6 +694,15 @@
       if (!equals(this[idx], other[idx])) return false;
     }
     return true;
+  }
+
+  //  Array$prototype$lte :: Array a ~> Array a -> Boolean
+  function Array$prototype$lte(other) {
+    for (var idx = 0; true; idx += 1) {
+      if (idx === this.length) return true;
+      if (idx === other.length) return false;
+      if (!equals(this[idx], other[idx])) return lte(this[idx], other[idx]);
+    }
   }
 
   //  Array$prototype$concat :: Array a ~> Array a -> Array a
@@ -711,6 +774,11 @@
     return Array$prototype$equals.call(this, other);
   }
 
+  //  Arguments$prototype$lte :: Arguments ~> Arguments -> Boolean
+  function Arguments$prototype$lte(other) {
+    return Array$prototype$lte.call(this, other);
+  }
+
   //  Error$prototype$toString :: Error ~> () -> String
   function Error$prototype$toString() {
     return 'new ' + this.name + '(' + toString(this.message) + ')';
@@ -749,6 +817,21 @@
     var keys = Object.keys(this).sort();
     return equals(keys, Object.keys(other).sort()) &&
            keys.every(function(k) { return equals(self[k], other[k]); });
+  }
+
+  //  Object$prototype$lte :: StrMap a ~> StrMap a -> Boolean
+  function Object$prototype$lte(other) {
+    var theseKeys = Object.keys(this).sort();
+    var otherKeys = Object.keys(other).sort();
+    while (true) {
+      if (theseKeys.length === 0) return true;
+      if (otherKeys.length === 0) return false;
+      var k = theseKeys.shift();
+      var z = otherKeys.shift();
+      if (k < z) return true;
+      if (k > z) return false;
+      if (!equals(this[k], other[k])) return lte(this[k], other[k]);
+    }
   }
 
   //  Object$prototype$concat :: StrMap a ~> StrMap a -> StrMap a
@@ -839,31 +922,36 @@
     Null: {
       prototype: {
         toString:                   Null$prototype$toString,
-        'fantasy-land/equals':      Null$prototype$equals
+        'fantasy-land/equals':      Null$prototype$equals,
+        'fantasy-land/lte':         Null$prototype$lte
       }
     },
     Undefined: {
       prototype: {
         toString:                   Undefined$prototype$toString,
-        'fantasy-land/equals':      Undefined$prototype$equals
+        'fantasy-land/equals':      Undefined$prototype$equals,
+        'fantasy-land/lte':         Undefined$prototype$lte
       }
     },
     Boolean: {
       prototype: {
         toString:                   Boolean$prototype$toString,
-        'fantasy-land/equals':      Boolean$prototype$equals
+        'fantasy-land/equals':      Boolean$prototype$equals,
+        'fantasy-land/lte':         Boolean$prototype$lte
       }
     },
     Number: {
       prototype: {
         toString:                   Number$prototype$toString,
-        'fantasy-land/equals':      Number$prototype$equals
+        'fantasy-land/equals':      Number$prototype$equals,
+        'fantasy-land/lte':         Number$prototype$lte
       }
     },
     Date: {
       prototype: {
         toString:                   Date$prototype$toString,
-        'fantasy-land/equals':      Date$prototype$equals
+        'fantasy-land/equals':      Date$prototype$equals,
+        'fantasy-land/lte':         Date$prototype$lte
       }
     },
     RegExp: {
@@ -876,6 +964,7 @@
       prototype: {
         toString:                   String$prototype$toString,
         'fantasy-land/equals':      String$prototype$equals,
+        'fantasy-land/lte':         String$prototype$lte,
         'fantasy-land/concat':      String$prototype$concat
       }
     },
@@ -887,6 +976,7 @@
       prototype: {
         toString:                   Array$prototype$toString,
         'fantasy-land/equals':      Array$prototype$equals,
+        'fantasy-land/lte':         Array$prototype$lte,
         'fantasy-land/concat':      Array$prototype$concat,
         'fantasy-land/map':         Array$prototype$map,
         'fantasy-land/ap':          Array$prototype$ap,
@@ -900,7 +990,8 @@
     Arguments: {
       prototype: {
         toString:                   Arguments$prototype$toString,
-        'fantasy-land/equals':      Arguments$prototype$equals
+        'fantasy-land/equals':      Arguments$prototype$equals,
+        'fantasy-land/lte':         Arguments$prototype$lte
       }
     },
     Error: {
@@ -915,6 +1006,7 @@
       prototype: {
         toString:                   Object$prototype$toString,
         'fantasy-land/equals':      Object$prototype$equals,
+        'fantasy-land/lte':         Object$prototype$lte,
         'fantasy-land/concat':      Object$prototype$concat,
         'fantasy-land/map':         Object$prototype$map,
         'fantasy-land/ap':          Object$prototype$ap,
@@ -1002,7 +1094,7 @@
   //.
   //. ```javascript
   //. > equals(0, -0)
-  //. false
+  //. true
   //.
   //. > equals(NaN, NaN)
   //. true
@@ -1018,7 +1110,7 @@
     var $pairs = [];
 
     return function equals(x, y) {
-      if (typeof x !== typeof y || type(x) !== type(y)) return false;
+      if (!sameType(x, y)) return false;
 
       //  This algorithm for comparing circular data structures was
       //  suggested in <http://stackoverflow.com/a/40622794/312785>.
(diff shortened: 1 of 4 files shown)


safareli:
do we need the `typeof` checks? I guess it's already checked in dispatcher part already

index.js
```diff
@@ -33,13 +33,13 @@
...
-//:               |               \ /     | | | |     \
-//:            Monoid         Traversable | | | |      \
-//:            (empty)        (traverse)  / | | \       \
+//:    |          |           \         / | | | | \
+//:    |          |            \       /  | | | |  \
+//:    |          |             \     /   | | | |   \
+//:    |          |              \   /    | | | |    \
+//:    |          |               \ /     | | | |     \
+//:   Ord      Monoid         Traversable | | | |      \
+//:  (lte)     (empty)        (traverse)  / | | \       \
```
davidchambers: Nit: For "even" words it appears that we're pointing to the character half a character to the right of centre (the second `o` in `Monoid`, for example). Let's make this line point to the `o` in `Setoid`. :)

index.js
```diff
@@ -255,6 +255,19 @@
...
+  //. `TypeClass` value for [Ord][].
+  //.
+  //. ```javascript
+  //. > Ord.test(0)
+  //. true
+  //.
+  //. > Ord.test(null)
+  //. true
+  //. ```
+  var Ord = $('Ord', [], {lte: Value});
```
davidchambers: The second argument should be `[Setoid]` to capture the dependency.

index.js
```diff
@@ -521,6 +544,11 @@
     return typeof other === typeof this && other.valueOf() === this.valueOf();
   }
 
+  //  Boolean$prototype$lte :: Boolean ~> Boolean -> Boolean
+  function Boolean$prototype$lte(other) {
+    return typeof other === typeof this && this.valueOf() <= other.valueOf();
+  }
+
```
davidchambers: In Sanctuary proper we decided not to work with Boolean, Number, and String objects. Since sanctuary-type-classes is essentially the standard library for Fantasy Land, we should not be so opinionated. As a result, we must accept that there are *four* Boolean values: `false`, `true`, `new Boolean(false)`, and `new Boolean(true)`. How should these values be ordered?

index.js
```diff
@@ -537,6 +565,16 @@
...
   }
 
+  //  Number$prototype$lte :: Number ~> Number -> Boolean
+  function Number$prototype$lte(other) {
+    return typeof other === 'object' ?
+      typeof this === 'object' &&
+      lte(this.valueOf(), other.valueOf()) :
+      isNaN(other) && isNaN(this) ||
+      typeof this !== 'object' &&
+      this <= other && (1 / this > 0 ? 1 : -1) <= (1 / other > 0 ? 1 : -1);
```
davidchambers: Could you explain `(1 / this > 0 ? 1 : -1) <= (1 / other > 0 ? 1 : -1)`?

index.js
```diff
@@ -640,6 +688,32 @@
     return true;
   }
 
+  //  Array$prototype$lte :: Array a ~> Array a -> Boolean
+  function Array$prototype$lte(other) {
+    for (var idx = 0; idx < this.length; idx += 1) {
+      if (other.length - idx === 0) {
```
davidchambers: This could be expressed more simply as `idx === other.length`, I believe.

test/index.js
```diff
@@ -522,6 +531,96 @@ test('equals', function() {
...
+  eq(Z.lte(false, false), true);
+  eq(Z.lte(false, true), true);
+  eq(Z.lte(true, false), false);
+  eq(Z.lte(true, true), true);
+  eq(Z.lte(new Boolean(false), new Boolean(false)), true);
+  eq(Z.lte(new Boolean(false), new Boolean(true)), true);
+  eq(Z.lte(new Boolean(true), new Boolean(false)), false);
+  eq(Z.lte(new Boolean(true), new Boolean(true)), true);
+  eq(Z.lte(false, new Boolean(false)), false);
+  eq(Z.lte(new Boolean(false), false), false);
```
davidchambers: This violates totality.

index.js
gabejohnson: We must either consider `new Boolean(false) === false` or consider them to be different types and not comparable. My implementations of `lte` return false if the values passed are of different types.

index.js
gabejohnson: Sorry. This expression was originally in another function and had an associated comment. The expression `(1 / a > 0 ? 1 : -1)` evaluates to the sign of `a`. It's used here to detect `-0`;

index.js
gabejohnson: Hahaha. I guess I was up too late when I wrote this. I'll fix it.

test/index.js
gabejohnson: Only if they are considered to be of the same type.

index.js
gabejohnson: If we do change it we'll also have to change the definition of `Boolean$prototype$equals`.

gabejohnson:
@safareli 
> I guess it's already checked in dispatcher part already

You are correct.

index.js
davidchambers: After giving this matter some thought, I suggest:

  - we consider `false` and `true` the two members of `BooleanPrimitive`;
  - we consider `new Boolean(false)` and `new Boolean(true)` the two members of `BooleanObject`; and
  - we have `Boolean$prototype$lte` define `fantasy-land/lte` for both `BooleanPrimitive` and `BooleanObject`.

Comparing a `BooleanPrimitive` to a `BooleanObject` (via `equals` or `lte`) will always result in `false`.

:thumbsup: / :thumbsdown: ?

gabejohnson: I understand the philosophy behind this thinking, but I don't see the immediate utility. Why might I want to distinguish between the boxed vs. primitive version of a value?

gabejohnson: Additionally, Sanctuary provides the `is` function and `is(Boolean, false) === true`. This is confusing if `equals(new Boolean(false), false) === false`.

davidchambers: I believe two values should be considered equal if and only if they can be used interchangeably. `false` and `new Boolean(false)` are not interchangeable, therefore they should not be considered equal. There are many situations in which these values are not interchangeable, but this is probably the most striking:

```javascript
> false ? '😵' : '🙂'
'🙂'

> new Boolean(false) ? '😵' : '🙂'
'😵'
```

gabejohnson: Boxed `Boolean`s are not properly coerced. WAT!!! I didn't realize.

index.js
```diff
@@ -583,6 +626,11 @@
     return typeof other === typeof this && other.valueOf() === this.valueOf();
   }
 
+  //  String$prototype$lte :: String ~> String -> Boolean
+  function String$prototype$lte(other) {
+    return typeof other === typeof this && this.valueOf() <= other.valueOf();
+  }
```
gabejohnson: @safareli, is this the `typeof` check you were referring to? If so, the one in `String$prototype$equals` would appear to be unnecessary as well.

davidchambers: If we decide to continue to treat `'abc'` and `new String('abc')` as different values, this `typeof` check will remain necessary. At the top level we ensure that `type(x) === type(y)`, but `type` doesn't differentiate between string primitives and String objects.

gabejohnson:
Updated

gabejohnson:
@davidchambers, does the behavior of `Object$prototype$lte` look correct to you?

test/index.js
```diff
@@ -186,6 +188,15 @@ test('Setoid', function() {
   eq(Z.Setoid.test({'@@type': 'my-package/Quux'}), true);
 });
 
+test('Ord', function() {
+  eq(type(Z.Ord), 'sanctuary-type-classes/TypeClass');
+  eq(Z.Ord.name, 'sanctuary-type-classes/Ord');
+  eq(Z.Ord.test(null), true);
+  eq(Z.Ord.test(''), true);
+  eq(Z.Ord.test([]), true);
+  // eq(Z.Ord.test({}), true);
```
davidchambers: Does this assertion now hold?

gabejohnson: Yes it does. I uncommented it and pushed the commit.

davidchambers:
> @davidchambers, does the behavior of `Object$prototype$lte` look correct to you?

Which is "larger": `{x: 0, y: 0, z: 0}` or `{x: 1}`? Why? Is there a reason to give structure (keys) precedence over values? We compare strings character by character rather than first comparing by length. I'm wondering whether there are two equally valid definitions, or whether there's a reason to favour one over the other.

gabejohnson:
By analogy with Arrays, I think doing the comparison with sorted keys makes sense.

`lte({x: 0, y: 0, z: 0}, {x: 1}) === true` because `{x: 0, y: 0, z: 0}` is equivalent to `[['x', 0], ['y', 0], ['z', 0]]` and `{x: 1}` is equivalent to `[['x', 1]]`.

I don't know that `{x: 0, y: 0, z: 0}` is comparable to `{w: 1}` however (therefore returning false).

My implementation, however, doesn't capture that behavior.

gabejohnson:
I've changed the behavior to match that of comparing a list of pairs sorted by key. If an object doesn't contain a key, its corresponding list contains a `['key', undefined]` pair.

index.js
davidchambers: Let's discuss and merge #47, then rebase this pull request and #45. :)

davidchambers:
@gabejohnson, please rebase this branch now that #47 has been merged.

gabejohnson:
Rebased

index.js
```diff
@@ -523,6 +546,11 @@
       this === other;
   }
 
+  //  Boolean$prototype$lte :: Boolean ~> Boolean -> Boolean
+  function Boolean$prototype$lte(other) {
+    return typeof other === typeof this && this <= other;
+  }
+
```
davidchambers: `typeof other === typeof this` is now a given as a result of #47. Let's use this approach:

```javascript
//  Boolean$prototype$lte :: Boolean ~> Boolean -> Boolean
function Boolean$prototype$lte(other) {
  return typeof this === 'object' ?
    lte(this.valueOf(), other.valueOf()) :
    this <= other;
```

I'm not happy with the type coercion in `this <= other`. I believe `!this || other` is equivalent.

gabejohnson: That's fine. Won't this return `true` for `lte(false, new Boolean(false))`? Is that where we came down on this?
```js
lte(false, new Boolean(false)) // true
lte(0, -0) // true
lte('a', new String('a')) // false
```

davidchambers: My comment was inaccurate. I'd forgotten the details of #47. What I suggest is that we update the `lte` function to check `typeof x === typeof y` in addition to `type(x) === type(y)`. We'll then be able to update `Boolean$prototype$lte`, `Number$prototype$lte`, and `String$prototype$lte`.

gabejohnson: Would you like me to make `equals(-0, 0) === true` as well?

davidchambers: > Would you like me to make `equals(-0, 0) === true` as well?

Yes. I see that you have already done so. :)

index.js
```diff
@@ -534,7 +564,14 @@
   function Number$prototype$equals(other) {
     return typeof this === 'object' ?
       equals(this.valueOf(), other.valueOf()) :
-      isNaN(this) && isNaN(other) || this === other && 1 / this === 1 / other;
+      isNaN(this) && isNaN(other) || this === other;
+  }
+
+  //  Number$prototype$lte :: Number ~> Number -> Boolean
+  function Number$prototype$lte(other) {
+    return typeof other === 'object' ?
```
davidchambers: Let's check `typeof this` as we do in `Number$prototype$equals`.

gabejohnson: Fixed

index.js
```diff
@@ -534,7 +564,14 @@
...
       equals(this.valueOf(), other.valueOf()) :
-      isNaN(this) && isNaN(other) || this === other && 1 / this === 1 / other;
+      isNaN(this) && isNaN(other) || this === other;
+  }
+
+  //  Number$prototype$lte :: Number ~> Number -> Boolean
+  function Number$prototype$lte(other) {
+    return typeof this === 'object' ?
+      lte(this.valueOf(), other.valueOf()) :
+      isNaN(other) && isNaN(this) || this <= other;
```
davidchambers: We could replace this with `this < other || equals(this, other)`. It reads nicely and avoids repeating the `isNaN` checks, though it would be less efficient. What do you think, Gabe?

If we decide to stick with the current approach, let's switch the `isNaN` checks to match `Number$prototype$equals`.

index.js
```diff
@@ -642,6 +691,27 @@
...
+      // the call to equals could have been expensive. A call to lte will be
+      // just as expensive.
+      // With compare it would be:
+      // try {
+      //   var cmp = compare(this[idx], other[idx]);
+      //   if (cmp === 'LT' || cmp === 'EQ') return true;
+      //   if (cmp === 'GT') return false;
+      // } catch() { return false; }
+      if (!equals(a, b)) {
+        return a === void 0 ? true : b === void 0 ? false : lte(a, b);
```
davidchambers: `xs[idx] === undefined` is not a reliable way to determine an array's bounds. It introduced a bug, in this case:

```javascript
> Z.lte([], [undefined])
true

> Z.lte([undefined], [])
true
```

Let's reinstate the `length` checks.

index.js
```diff
@@ -751,6 +826,30 @@
            keys.every(function(k) { return equals(self[k], other[k]); });
   }
 
+  function unionKeys(a, b) {
+    var keys = Object.keys(a);
+    var bKeys = Object.keys(b);
+    for (var i = 0; i < bKeys.length; i += 1) {
+      if (a[bKeys[i]] === void 0) keys.push(bKeys[i]);
```
davidchambers: This `=== undefined` check looks suspect to me. Let's assume `a` is `{}` and `b` is `{toString: () => 'x'}`. This means `keys` will be `[]` initially and `bKeys` will be `['toString']`. `a[bKeys[0]]` will not evaluate to `undefined` because `a['toString']` evaluates to `Object.prototype.toString`.

I think the check should be `!Object.prototype.hasOwnProperty.call(a, bKeys[i])`.

This approach might be clearer:

```javascript
var key;
var keys = {};
for (key in a) keys[key] = 0;
for (key in b) keys[key] = 0;
return Object.keys(keys);
```

index.js
```diff
@@ -1035,6 +1143,41 @@
...
+  //.
+  //. > lte([1, 2], [3, 4])
+  //. true
+  //.
+  //. > lte([1, 3], [1, 2])
+  //. false
+  //. ```
+  function lte(x, y) {
+    return (typeof x === typeof y && type(x) === type(y))
+      && Ord.methods.lte(x)(y);
```
davidchambers: I'd prefer this formatting:

```javascript
return typeof x === typeof y &&
       type(x) === type(y) &&
       Ord.methods.lte(x)(y);
```

test/index.js
```diff
@@ -524,6 +535,110 @@ test('equals', function() {
...
+  eq(Z.lte('abc', new String('abc')), false);
+  eq(Z.lte(new String('abc'), 'abc'), false);
+  eq(Z.lte([], []), true);
+  eq(Z.lte([1, 2], [1, 2]), true);
+  eq(Z.lte([1, 2, 3], [1, 2]), false);
+  eq(Z.lte([1, 2], [1, 2, 3]), true);
+  eq(Z.lte([1, 2], [2]), true);
+  /* eslint-disable comma-spacing */
+  eq(Z.lte([1,,], [1, 0]), true); // eslint-disable-line no-sparse-arrays
+  /* eslint-enable comma-spacing */
```
davidchambers: I think we could use `// eslint-disable-next-line comma-spacing no-sparse-arrays` on the line preceding the `[1,,]` to avoid having comments on three consecutive lines. ;)

index.js
```diff
@@ -1035,6 +1143,41 @@
...
+  //. `fantasy-land/lte` implementations are provided for the following
+  //. built-in types: Null, Undefined, Boolean, Number, Date, String,
+  //. Array, Arguments, Error, and Object.
+  //.
+  //. The algorithm supports circular data structures. Two arrays are lte
+  //. if they have the same index paths and for each path have lte values.
+  //. An array which represents `[1, [1, [1, [1, [1, ...]]]]]` and one that
+  //. represents `[2, [2, [2, [2, [2, ...]]]]]`, for example,
+  //. are lte even if their internal structures differ. Two objects are lte
+  //. if they have the same property paths and for each path have lte values.
```
davidchambers: Let's shorten this paragraph to something like this:

_The algorithm supports circular data structures in the same manner as [`equals`][1]._


[1]: https://github.com/sanctuary-js/sanctuary-type-classes#equals

index.js
gabejohnson: At this low of a level, I'd opt for efficiency. I do agree with fixing the `isNaN` though.

test/index.js
gabejohnson: Ahhhh...yes. I just learned about that one a couple of days ago. Agreed.

index.js
gabejohnson: You're right about the check being bogus. I guess the question then is should we consider _enumerable_ properties, _own_ properties, both or neither. When we're talking about a `StrMap` are we only interested in _own_ properties?

test/index.js
gabejohnson: I'm not even sure that this is a valid test. What is the type of `[1,,]`? Surely not `Array Number`?

davidchambers: Let's remove this test case. :)

index.js
davidchambers: Let's follow the precedent set by `Object$prototype$map`, which uses an unguarded `for...in` loop. In other words, let's respect enumerable properties regardless of where they're "attached".

gabejohnson:
Updated

index.js
```diff
@@ -751,6 +827,32 @@
...
+    for (var idx = 0; idx < keys.length; idx += 1) {
+      var key = keys[idx];
+      var a = this[key], b = other[key];
+
+      if (!equals(a, b)) {
+        var inThis = Object.prototype.propertyIsEnumerable.call(this, key);
+        var inOther = Object.prototype.propertyIsEnumerable.call(other, key);
+        return  inThis === false ? true :
+                inOther === false ? false :
+                lte(a, b);
```
davidchambers: Let's simplify this to `!inThis || inOther && lte(a, b)`.

index.js
```diff
@@ -642,6 +691,28 @@
     return true;
   }
 
+  //  Array$prototype$lte :: Array a ~> Array a -> Boolean
+  function Array$prototype$lte(other) {
+    for (var idx = 0; idx < this.length; idx += 1) {
+      if (other.length === idx) return false;
```
davidchambers: This is a Yoda condition. Please switch the operands. :)

index.js
```diff
@@ -751,6 +827,32 @@
...
+        var inOther = Object.prototype.propertyIsEnumerable.call(other, key);
+        return  inThis === false ? true :
+                inOther === false ? false :
+                lte(a, b);
+      }
+    }
+
+    return true;
+  }
+
```
davidchambers: I came up with a different implementation which I find elegant:

```javascript
//  Object$prototype$lte :: StrMap a ~> StrMap a -> Boolean
function Object$prototype$lte(other) {
  var theseKeys = Object.keys(this).sort();
  var otherKeys = Object.keys(other).sort();
  while (true) {
    if (theseKeys.length === 0) return true;
    if (otherKeys.length === 0) return false;
    var k = theseKeys.shift();
    var z = otherKeys.shift();
    if (k < z) return true;
    if (k > z) return false;
    if (!equals(this[k], other[k])) return lte(this[k], other[k]);
  }
}
```

davidchambers: Here's a matching implementation of `Array#fantasy-land/lte`:

```javascript
//  Array$prototype$lte :: Array a ~> Array a -> Boolean
function Array$prototype$lte(other) {
  for (var idx = 0; true; idx += 1) {
    if (idx === this.length) return true;
    if (idx === other.length) return false;
    if (!equals(this[idx], other[idx])) return lte(this[idx], other[idx]);
  }
}
```

gabejohnson: I like that they're much cleaner. The implementation of `Object$prototype$lte` changes the semantics though. Previously, `Z.lte({x: 0, y: 0, z: 0}, {y: 0, z: 1}) === false` as this was conceptually testing:
```js
Z.lte([0, 0, 0], [undefined, 0, 1])
```
See https://github.com/sanctuary-js/sanctuary-type-classes/pull/46#issuecomment-293663721

davidchambers: Defaulting to `undefined` bothered me, and I just realized it's not lawful:

```javascript
> Z.lte({}, {x: undefined})
true

> Z.lte({x: undefined}, {})
true

> Z.equals({}, {x: undefined})
false
```

gabejohnson: Changing:
```js
if (k < z) return true;
if (k > z) return false;
```
to
```js
if (k !== z) return false;
```
should fix this.

gabejohnson: I was typing that up while you responded.

gabejohnson: We should define the semantics precisely before proceeding.

gabejohnson: Looking at [Data.StringMap](http://hackage.haskell.org/package/data-stringmap-1.0.1.1/docs/Data-StringMap.html) would be a good start.

index.js
```diff
@@ -1035,6 +1122,108 @@
...
+  //. false
+  //.
+  //. > lt(0, 1)
+  //. true
+  //.
+  //. > lt(1, 0)
+  //. false
+  //. ```
+  function lt(x, y) {
+    return !lte(y, x);
```
gabejohnson: Won't this return `true` if `x` and `y` are of different types?
```js
lte(0, '0'); // false
lt('0', 0); // true
```

index.js
```diff
@@ -1035,6 +1122,108 @@
...
+  //. false
+  //.
+  //. > gt(0, 1)
+  //. false
+  //.
+  //. > gt(1, 0)
+  //. true
+  //. ```
+  function gt(x, y) {
+    return !lte(x, y);
```
gabejohnson: Same issue as with `lt`.

index.js
davidchambers: Good catch! Fixed. :)

davidchambers:
This is now close, I think! Please have a look. :)

index.js
```diff
@@ -605,6 +605,11 @@
            other.unicode === this.unicode;
   }
 
+  //  RegExp$prototype$lte :: RegExp ~> RegExp -> Boolean
+  function RegExp$prototype$lte(other) {
+    return lte(toString(this), toString(other));
+  }
```
gabejohnson: @davidchambers I had originally implemented `lte` for `RegExp` as well but couldn't come up with a satisfying semantic for ordering on flags. Just doing `String` comparison seems like it is avoiding the issue.

davidchambers: Do you think a string comparison is reasonable, or are you unhappy with the way it smushes source and flags together?

Another option would be to sort `/abc/` as `['', 'abc']` and `/def/g` as `['g', 'def']`.

gabejohnson: I think string comparison is reasonable for _exact_ match patterns like `/abc/`. I'm not sure that the idea of comparing `/\b([01]+b|\d+|[\da-f]+h)\b/` with something like `/\b\d+ (pig|cow|chicken)s?\b/` makes sense though.

Also, I don't see the immediate utility.

davidchambers: > I don't see the immediate utility.

A binary search tree needs to compare values during insertion and lookup. It seems reasonable to expect to be able to store RegExp objects in such a tree.

gabejohnson: I would argue to use a binary tree that takes a custom comparison function in this case, but I won't argue too strongly 😄

gabejohnson:
@davidchambers are you waiting on anything before merging?

index.js
```diff
@@ -711,6 +779,11 @@
     return Array$prototype$equals.call(this, other);
   }
 
+  //  Arguments$prototype$lte :: Arguments ~> Arguments -> Boolean
+  function Arguments$prototype$lte(other) {
+    return Array$prototype$lte.call(this, other);
+  }
+
   //  Error$prototype$toString :: Error ~> () -> String
   function Error$prototype$toString() {
```
safareli: Maybe add lte for Error too?

gabejohnson: I had an implementation that I removed. @safareli would you compare `Error.prototype.name` and `Error.prototype.message`? Or just use `toString`?

safareli: hm, sort with name and then message?

safareli: or it could be done in another PR if someone needs it

gabejohnson: I think we could wait. I feel the same way about `RegExp`.

@davidchambers what are your thoughts?

davidchambers: Let's remove `RegExp$prototype$lte` for now. As Irakli rightly pointed out, we can easily add `fantasy-land/lte` implementations for Error and RegExp at some point in the future.

davidchambers:
I've squashed. I'm giving this a final perusal now. :)

PR #47, opened 2017-04-13
perform typeof check in ‘equals’

davidchambers:
Commit message:

>     This simplifies Number$prototype$equals (and the corresponding Boolean
>     and String functions) by reducing the number of permutations from four
>     to two (indicated by *):
>
>         typeof this     typeof other
>         --------------------------------
>       * 'object'        'object'
>         'object'        'number'
>         'number'        'object'
>       * 'number'        'number'
>
>     This commit also fixes a bug which caused Z.equals(new Number(NaN), NaN)
>     to evaluate true rather than false.

### DIFF
--- a/index.js
+++ b/index.js
@@ -518,7 +518,9 @@
 
   //  Boolean$prototype$equals :: Boolean ~> Boolean -> Boolean
   function Boolean$prototype$equals(other) {
-    return typeof other === typeof this && other.valueOf() === this.valueOf();
+    return typeof this === 'object' ?
+      equals(this.valueOf(), other.valueOf()) :
+      this === other;
   }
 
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -518,7 +518,9 @@
 
   //  Boolean$prototype$equals :: Boolean ~> Boolean -> Boolean
   function Boolean$prototype$equals(other) {
-    return typeof other === typeof this && other.valueOf() === this.valueOf();
+    return typeof this === 'object' ?
+      equals(this.valueOf(), other.valueOf()) :
+      this === other;
   }
```
davidchambers: In all three functions the `'object'` branch is simply `equals(this.valueOf(), other.valueOf())`. This consistency pleases me. :)

PR #50, opened 2017-05-02
ensure consistent enumeration order in Object$prototype$reduce

davidchambers:
See sanctuary-js/sanctuary#389 for the motivation behind this change.

### DIFF
--- a/index.js
+++ b/index.js
@@ -861,9 +861,9 @@
 
   //  Object$prototype$reduce :: StrMap a ~> ((b, a) -> b, b) -> b
   function Object$prototype$reduce(f, initial) {
-    var result = initial;
-    for (var k in this) result = f(result, this[k]);
-    return result;
+    var self = this;
+    function reducer(acc, k) { return f(acc, self[k]); }
+    return Object.keys(this).sort().reduce(reducer, initial);
   }
 
   //  Function$of :: b -> (a -> b)
--- a/test/index.js
+++ b/test/index.js
@@ -930,6 +930,8 @@ test('reduce', function() {
   eq(Z.reduce(Z.concat, 'x', ['a', 'b', 'c']), 'xabc');
   eq(Z.reduce(add, 0, {}), 0);
   eq(Z.reduce(add, 0, {a: 1, b: 2, c: 3, d: 4, e: 5}), 15);
+  eq(Z.reduce(function(xs, x) { return Z.concat(xs, [x]); }, [], {a: 1, b: 2, c: 3}), [1, 2, 3]);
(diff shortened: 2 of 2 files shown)


gabejohnson:
How about this?
```js
//  Object$prototype$reduce :: StrMap a ~> ((b, a) -> b, b) -> b
function Object$prototype$reduce(f, initial) {
  var result = initial;
  var keys = Object.keys(this).sort();
  for (var i = 0; i < keys.length; i += 1) result = f(result, this[keys[i]]);
  return result;
}
```
It cuts out the `map` traversal.

kedashoe:
This still makes me nervous in general. First of all, the problem at hand, calling a function like `find` on an unordered collection and getting different results.. has this ever actually bitten anyone? In the toy example from sanctuary-js/sanctuary#389 sure you might get different results.. but when would you ever do something like that? And in the new tests, we have exclamation points when we find 4, but that seems totally reasonable to me when looking for an even number in an unordered collection: you say grab me an even number from this bag and I say sure, here is 4!

Fixing it in this way has the negative impact of slowing down every other reduce operation on objects. I feel like the solution here is to provide an alternative `find` with an `Ord` constraint on the container or something like that? With this PR we are adding a Foldable constraint but also a hidden Ord constraint.

This particular case aside, I'm a bit troubled with these general solutions when we can provide very simple and more efficient ones. For example, `filter`. If we call sanctuary's filter on an array of 10,000 elements and 8,000 satisfy the predicate, we create 16,000 intermediate arrays when we don't need any! And if we do want `find` to be ordered for Objects, we can also do that in a single pass.

Currently types are only allowed to provide their own implementation if there is an associated fantasy land definition. I think it would be great to have some way to dispatch to some of these other functions as well. I'm not sure if `find` will ever have a fantasy land definition? Something like `filter`, we can see that we need to be an Applicative and a Foldable, so can we have some mechanism to dispatch to `filter` for types that satisfy these constraints and have a `filter` function? We could even use the sanctuary namespace if we wanted, eg 'sanctuary/filter' for dispatching. Could provide laws that users could test against if they want. Seems like it would be nice for our users to at least provide these more optimal implementations for built in types, and in the long run a way where they can do the same for their own.

gabejohnson:
@kedashoe, prefixed methods? `'sanctuary/filter`? Dispatch by default and derived otherwise?

gabejohnson:
That way we could test out/drive prospective algebras. For instance, https://github.com/fantasyland/fantasy-land/issues/208. Sanctuary could implement this under it's own namespace and just dispatch to the FL prefix if it ever gets added to the spec.

davidchambers:
Thanks for the suggestion, Gabe. I've updated the implementation. How does it look to you now?

gabejohnson:
Looks solid.

PR #55, opened 2017-05-17
fantasy-land@3.3.0

### LINKED ISSUE #54: fantasy-land@3.3.0
<https://github.com/fantasyland/fantasy-land/compare/v3.2.0...v3.3.0>

Feel free to assign this issue to yourself if you would like to work on it, @gabejohnson.

gabejohnson:
Closes #54

### DIFF
--- a/index.js
+++ b/index.js
@@ -31,37 +31,37 @@
 //. ## Type-class hierarchy
 //.
 //. <pre>
-//:  Setoid   Semigroup   Foldable        Functor      Contravariant
-//: (equals)   (concat)   (reduce)         (map)        (contramap)
-//:     |         |           \         / | | | | \
-//:     |         |            \       /  | | | |  \
-//:     |         |             \     /   | | | |   \
-//:     |         |              \   /    | | | |    \
-//:     |         |               \ /     | | | |     \
-//:    Ord     Monoid         Traversable | | | |      \
-//:   (lte)    (empty)        (traverse)  / | | \       \
-//:                                      /  | |  \       \
-//:                                     /   / \   \       \
-//:                             Profunctor /   \ Bifunctor \
-//:                              (promap) /     \ (bimap)   \
-//:                                      /       \           \
-//:                                     /         \           \
-//:                                   Alt        Apply      Extend
-//:                                  (alt)        (ap)     (extend)
-//:                                   /           / \           \
-//:                                  /           /   \           \
-//:                                 /           /     \           \
-//:                                /           /       \           \
-//:                               /           /         \           \
-//:                             Plus    Applicative    Chain      Comonad
-//:                            (zero)       (of)      (chain)    (extract)
-//:                               \         / \         / \
-//:                                \       /   \       /   \
-//:                                 \     /     \     /     \
-//:                                  \   /       \   /       \
-//:                                   \ /         \ /         \
-//:                               Alternative    Monad     ChainRec
-//:                                                       (chainRec)
+//:  Setoid   Semigroupoid  Semigroup   Foldable        Functor      Contravariant
+//: (equals)    (compose)    (concat)   (reduce)         (map)        (contramap)
+//:     |           |           |           \         / | | | | \
+//:     |           |           |            \       /  | | | |  \
+//:     |           |           |             \     /   | | | |   \
+//:     |           |           |              \   /    | | | |    \
+//:     |           |           |               \ /     | | | |     \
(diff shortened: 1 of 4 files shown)


gabejohnson:
The changes to the dependency diagram don't pass lint.

index.js
```diff
@@ -31,37 +31,37 @@
...
+//:                                             /            /          \           \
+//:                                           Plus     Applicative     Chain      Comonad
+//:                                          (zero)       (of)        (chain)     (extract)
+//:                                             \         /  \         /  \
+//:                                              \       /    \       /    \
+//:                                               \     /      \     /      \
+//:                                                \   /        \   /        \
+//:                                                 \ /          \ /          \
+//:                                              Alternative    Monad      ChainRec
+//:                                                                       (chainRec)
```
gabejohnson: I was attempting to fix the alignment of branches. I might leave the fiddling to you though David if that's okay.

index.js
```diff
@@ -875,6 +901,11 @@
     }, of(typeRep, {}));
   }
 
+  //  Function$id :: () -> a -> a
+  function Function$id() {
+    return function(x) { return x; };
```
Avaq: It may be nicer to return the same identify function every time, rather than creating a new identity function over and over again. Currently `S.id(Function) !== S.id(Function)`.

index.js
```diff
@@ -1256,6 +1295,39 @@
...
+  //# compose :: Semigroupoid a => (a, a) -> a
+  //.
+  //. Function wrapper for [`fantasy-land/compose`][].
+  //.
+  //. `fantasy-land/compose` implementations are provided for the following
+  //. built-in types: Function
+  //.
+  //. ```javascript
+  //. > concat('abc', 'def')
+  //. 'abcdef'
```
Avaq: I think this code example might be here by mistake.

index.js
gabejohnson: I thought about that. I just went off of what `empty` does. I suppose I could make one function and freeze it.

gabejohnson:
```js
> Z.equals(Z.id(Function), Z.id(Function));
false
```
Definitely have to change that.

Avaq: No need to freeze, just hoist:

```js
function identity(x) { return x; }
function Function$id() { return identity; }
```

index.js
```diff
@@ -104,6 +104,7 @@
 
   //  identity :: a -> a
   function identity(x) { return x; }
+  Object.freeze(identity);
```
Avaq: There's no need to freeze this object I think. Plus it may be increasing our minimum JS run-time requirements, but I'm not sure about that.

gabejohnson: If we're exposing a single instance of this function I'd like to make it "immutable".

`Object.freeze` is ES5. It's supported in IE9.

index.js
```diff
@@ -891,6 +923,12 @@
     };
   }
 
+  //  Function$prototype$compose:: (b -> c) ~> (a -> b) -> (a -> c)
```
gabejohnson: I'm not sure about this signature. Did I get it wrong in https://github.com/fantasyland/fantasy-land/pull/248?

davidchambers:
I hopped on a plane after suggesting you work on a pull request, Gabe, and found this ready for me to review upon landing. Exciting! :airplane:

index.js
davidchambers: Fiddling successful. :)

index.js
davidchambers: I'm anti-freeze too. What are you thinking could happen if we do not freeze this function, Gabe? Someone might attach properties to it? Would this even cause problems?

Currently we don't even freeze `Z` itself, and mutation such as `Z.lt = Z.gt;` could certainly cause problems in practice. ;)

gabejohnson: I concede. If a lib/user wants to add properties to `identity` they're doing it wrong.
> Would this even cause problems?

Probably not. I'm just being overly cautious.

gabejohnson:
@davidchambers I'd like to get https://github.com/fantasyland/fantasy-land/pull/249 resolved before merging this.

gabejohnson:
I've closed fantasyland/fantasy-land#249 leaving the ordering unchanged. The implementation of `Function$prototype$compose` now matches.

davidchambers:
This is looking good, Gabe. Please squash the commits. :)

gabejohnson:
Done

davidchambers:
Thanks, Gabe!

PR #56, opened 2017-05-19
The result of a session of meditative code untangling

Avaq:
I originally wanted to do some performance optimizations, but found that I needed to untangle the code before I could understand what's going on. Instead of performance optimizations, I decided to simply commit the untangled code, which should be easier to optimize.

#### Separate _funcPath(true) and _funcPath(false)

I consider functions where any number of leading Booleans determine the behavior somewhat of an anti-pattern of abstraction. Separating the two behaviors to their depending functions greatly improves readability.

#### Separate getBoundMethod for Value and for Constructor

Pretty much the same story as what's described above. We already know beforehand which of the internal behaviors of `getBoundMethod` we want to run, but we still route our code through this wrapper which determines it *again*. Getting rid of the wrapper allowed me to hoist the functions, which improves performance and readability.

#### ~Simplify type class creation API~

~After the previous changes were made, `StrMap Location`-structure was immediately broken into two separate `Array String`s. All that conversion code could be removed by simply passing it into the function "un-encoded".~


index.js
```diff
@@ -1301,7 +1301,7 @@
   //. Nil
   //. ```
   function empty(typeRep) {
-    return Monoid.methods.empty(typeRep)();
+    return Monoid.functions.empty(typeRep)();
```
Avaq: You may not approve of this change, since suddenly function definitions need to know where the implementation lives. If that's a problem we can simply add something along the lines of...:

```js
typeClass.mixed = typeClass.functions.concat(typeClass.methods);
```

...at the end of `$`, and use the `mixed` property to access implementations.

davidchambers:
I like the title of this pull request. :)

The motivation for these changes is unclear. Could you update the description or comment below?

Avaq:
I added a motivation for each commit to the PR description

index.js
```diff
@@ -172,80 +172,80 @@
...
-      if (x == null || !(allowInheritedProps || has(k, x))) return null;
-      x = x[k];
+  //  funcPath :: (Array String, a) -> Nullable Function
+  function funcPath(path, value) {
+    var x = value;
+    for (var i = 0; i < path.length; i += 1) {
+      x = x && x[path[i]];
+      if (!x) return null;
     }
     return typeof x === 'function' ? x : null;
```
Avaq: This is one of the few changed lines, but after the simplification of the algorithm, the `null`-branch was never taken. I added this `Bad()`-structure to the tests in order to create the input necessary to reach this branch. If you think the `Bad()` input should actually just lead to a TypeError then we can remove the ternary and the new test cases.

index.js
```diff
@@ -172,80 +172,80 @@
...
       });
     });
 
-    typeClass.methods = keys.reduce(function(methods, _name) {
-      methods[_name] = getBoundMethod(_name);
-      return methods;
-    }, {});
+    functionKeys.forEach(function(k) {
+      this[k] = getBoundFunction(k);
+    }, typeClass.functions = {});
```
Avaq: I prefer this slightly over a mutating `reduce`. At least when I see `forEach` I can be sure that intentional mutation is happening there.

davidchambers:
```diff
-  var Setoid = $('Setoid', [], {equals: Value});
+  var Setoid = $('Setoid', [], [], ['equals']);
```

I appreciate your desire to improve performance and reduce implementation complexity, but I find the above change concerning for several reasons:

  - we have three arrays and it's not obvious whether any of them are related (this resembles the Boolean trap);
  - `{equals: Constructor}` would stand out to me as an error, but `['equals'], []` would be difficult to spot (this is an instance of the ordering problem);
  - invalid specifications such as `['equals'], ['equals']` are representable; and
  - we use an ordered data structure (`Array String`) to represent values with no inherent ordering (the method names).

Avaq:
> I find the [simplification of the class creation API] concerning

You put forth valid points, and if you keep your position I will revert the associated commit. But first I want to make sure that you've asked yourself whether it's worth the [17 lines of added complexity](https://github.com/sanctuary-js/sanctuary-type-classes/pull/56/commits/f77c014b3dffc638c80cf995990fedfce4a2cb3b) to improve an API which is essentially just an internal mechanism for code-reuse. Furthermore `['equals'], []` would fail the unit tests, so that point is only partly valid. ;)

Avaq:
Rebased.

davidchambers:
> I want to make sure that you've asked yourself whether it's worth the 17 lines of added complexity to improve an API which is essentially just an internal mechanism for code-reuse.

There may be 17 lines of code, but they're 17 straightforward lines of code. These lines are about as simple as one can imagine:

```javascript
//  Constructor :: Location
var Constructor = 'Constructor';
```

Perhaps you're unhappy with this code:

```javascript
var keys = Object.keys(requirements);

var functionKeys = keys.filter(function(k) {
  return requirements[k] === Constructor;
});

var methodKeys = keys.filter(function(k) {
  return requirements[k] === Value;
});
```

We could do something like this instead:

```javascript
var functionKeys = [];
var methodKeys = [];
for (var name in requirements) {
  (requirements[name] === Value ? methodKeys : functionKeys).push(name);
}
```

> You put forth valid points, and if you keep your position I will revert the associated commit.

Please revert the associated commit. I'm happy to revisit this matter if you feel strongly about it, but I'd like to first review and merge your other changes.

index.js
Avaq: I could move `typeClass.functions = {}` to above, and replace `this` by `typeClass.functions` for readability.

davidchambers: I would find that clearer. :thumbsup:

index.js
```diff
@@ -180,72 +180,89 @@
...
-  }
+  //  Prefix :: String
+  var Prefix = 'fantasy-land/';
 
   //  funcPath :: (Array String, a) -> Nullable Function
-  function funcPath(path, x) {
-    return _funcPath(true, path, x);
+  function funcPath(path, value) {
+    var x = value;
+    for (var i = 0; i < path.length; i += 1) {
```
davidchambers: `:%s/\<i\>/idx/g`

index.js
```diff
@@ -180,72 +180,91 @@
...
+
+  //  funcPath :: (Array String, a) -> Nullable Function
+  function funcPath(path, value) {
+    var x = value;
     for (var idx = 0; idx < path.length; idx += 1) {
-      var k = path[idx];
-      if (x == null || !(allowInheritedProps || has(k, x))) return null;
-      x = x[k];
+      x = x && x[path[idx]];
+      if (!x) return null;
```
davidchambers: Could we move this to the test in the `for` construct? Also, could we avoid the type coercion?

index.js
```diff
@@ -180,72 +180,91 @@
...
+    for (var idx = 0; idx < path.length; idx += 1) {
+      x = has(path[idx], x) && x[path[idx]];
+      if (!x) return null;
+    }
+    return x;
   }
 
   //  functionName :: Function -> String
-  var functionName = 'name' in function f() {} ?
+  var functionName = 'name' in identity ?
```
davidchambers: I find it clearer to define a function in situ for this check.

index.js
```diff
@@ -180,72 +180,91 @@
...
     function functionName(f) { return f.name; } :
     /* istanbul ignore next */
     function functionName(f) {
       var match = /function (\w*)/.exec(f);
       return match == null ? '' : match[1];
     };
 
+  function getBoundFunction(_name) {
```
davidchambers: Perhaps this function should be named `getFunction`, since nothing is being bound.

index.js
davidchambers: I have no qualms about specifying the location of the implementation here. :)

index.js
Avaq: But then the code coverage complains that the function is never called.

index.js
Avaq: I had the same thought, though `getFunction` seems a bit general

index.js
```diff
@@ -180,72 +180,91 @@
...
+
     var typeClass = TypeClass(name, dependencies, function(x) {
-      return keys.every(function(_name) {
-        var arg = requirements[_name] === Constructor ? x.constructor : x;
-        return getBoundMethod(_name)(arg) != null;
+      return functionKeys.every(function(k) {
+        return typeClass.functions[k](x.constructor) != null;
+      }) && methodKeys.every(function(k) {
+        return typeClass.methods[k](x) != null;
       });
```
davidchambers: I'd prefer this formatting:

```javascript
return functionKeys.every(function(k) {
         return typeClass.functions[k](x.constructor) != null;
       }) &&
       methodKeys.every(function(k) {
         return typeClass.methods[k](x) != null;
       });
```

index.js
```diff
@@ -180,72 +180,89 @@
...
-    return typeof x === 'function' ? x : null;
-  }
+  //  Prefix :: String
+  var Prefix = 'fantasy-land/';
 
   //  funcPath :: (Array String, a) -> Nullable Function
-  function funcPath(path, x) {
-    return _funcPath(true, path, x);
+  function funcPath(path, value) {
+    var x = value, len = path.length;
```
davidchambers: We don't cache `length` in other functions. I'm open to making this change for all such occurrences in a separate pull request. For this pull request, let's remove `len` from this function and from `implPath`.

index.js
Avaq: But we have so many fine functions just laying around! :(

index.js
```diff
@@ -180,72 +180,89 @@
...
-    var x = _x;
-    for (var idx = 0; idx < path.length; idx += 1) {
-      var k = path[idx];
-      if (x == null || !(allowInheritedProps || has(k, x))) return null;
-      x = x[k];
-    }
-    return typeof x === 'function' ? x : null;
-  }
+  //  Prefix :: String
+  var Prefix = 'fantasy-land/';
```
davidchambers: I would prefer a lower-case "p" here, since this is a member of `String` rather than a data constructor (`Constructor` and `Value` are intended to be viewed as the data constructors of the `Location` type).

index.js
davidchambers: I don't think adding a misleading word to a function's name is a good way to make a name more specific. :P

index.js
davidchambers: We do, but I don't like coupling this check to any one of them.

If you *really* want to get rid of `function f() {}`, I could get behind replacing it with `Function`. :)

index.js
Avaq: Ah. Whoops, that's a left-over of me trying to get the line below on a one-line.

index.js
```diff
@@ -180,72 +180,89 @@
...
     var typeClass = TypeClass(name, dependencies, function(x) {
-      return keys.every(function(_name) {
-        var arg = requirements[_name] === Constructor ? x.constructor : x;
-        return getBoundMethod(_name)(arg) != null;
-      });
+      return functionKeys.every(function(k) {
+               return typeClass.functions[k](x.constructor) != null;
+             }) && methodKeys.every(function(k) {
+               return typeClass.methods[k](x) != null;
+             });
```
Avaq: `eslint indent` does not approve.

davidchambers: Let's add `--rule 'indent: [off]'` to the makefile. `indent` is one rule I'm considering disabling in sanctuary-style as it produces so many false positives.

index.js
Avaq: Ahhh, that's amazing!

index.js
gabejohnson: I'm fairly certain `length` caching is unnecessary in general. Modern browsers do it themselves.

index.js
Avaq: I suppose I could get behind that.

Avaq:
GitHub says @gabejohnson left a review, but I can't find it anywhere.

gabejohnson:
Meh. It was just a comment on an outdated diff https://github.com/sanctuary-js/sanctuary-type-classes/pull/56#discussion_r117506931

test/Bad.js
```diff
@@ -0,0 +1,30 @@
...
+Bad.prototype[FL.reduce] = 'nonsense';
+Bad.prototype[FL.traverse] = 'nonsense';
+Bad.prototype[FL.extend] = 'nonsense';
+Bad.prototype[FL.extract] = 'nonsense';
+Bad.prototype.inspect =
+Bad.prototype.toString = function() {
+  return 'Bad()';
+};
+
+module.exports = Bad;
```
davidchambers: It appears this file is no longer necessary. :)

index.js
```diff
@@ -180,72 +180,92 @@
...
-  var functionName = 'name' in function f() {} ?
+  var functionName = 'name' in Function ?
     function functionName(f) { return f.name; } :
     /* istanbul ignore next */
     function functionName(f) {
       var match = /function (\w*)/.exec(f);
       return match == null ? '' : match[1];
     };
 
+  //  getFunction :: String -> TypeRep -> Nullable Function
```
davidchambers: `s/TypeRep/TypeRep a/`

index.js
```diff
@@ -180,72 +180,92 @@
...
+    var name = prefix + _name;
+    return function(typeRep) {
+      var f = funcPath([name], typeRep);
+      return f == null && typeof typeRep === 'function' ?
+        implPath([functionName(typeRep), name]) :
+        f;
+    };
+  }
+
+  //  getBoundMethod :: String -> a -> Nullable (a ~> b)
```
davidchambers: `(a ~> b)` is cute, but inaccurate. The resulting function may be called with no arguments (since the `a` has already been bound). I think `Nullable Function` is as accurate as we can make the return type.

davidchambers:
I believe the final section in the description is no longer applicable and should be `<del>`eted.

index.js
Avaq: I think since we always represent a nullary *methods* as `a ~> () -> b`, this signature is valid. Because `() -> b` is a member of `b`? We could probably even make it more precise by changing it to `a ~> Function`.

Come to think of it, the `~>` is very strange. I'm currently using it like a "return" arrow, but it's not quite that. It shouldn't have a return type unless we allow `a ~> b` to represent a nullary method. It's currently completely meaningless if not followed by an argument list. It's actually more like the `~> ()`-operator. So in other words: I'm using a made-up operator, and I will cease to do so immediately.

---

### Some rambling

Why don't we allow `a ~> b`? I think we can still represent everything we want with it:

```hs
f :: a ~> b -- nullary method returning anything
g :: a ~> (b, c) -> d -- binary method returning anything
h :: a ~> ((b, c) -> d) -- nullary method on A returning a binary function
i :: a ~> Function -- a method of unknown arity
j :: a ~> (Function) -- a method returning a function of unknown arity
bind :: a -> Function -> (a ~> Function) -- binary function returning a method with unknown arity
```

`h` and `j` are very strange, but I think we already have this problem in the way we represent ramda-style curried functions vs actual curried functions:

```hs
f :: a -> b -> c -- we can call f(a, b)
g :: a -> (b -> c) -- we must call g(a)(b)
```

davidchambers:
If you squash, I will merge. :)

Avaq:
I couldn't really think of a single description for a commit, so I went with my two original commits, plus an additional one for the change from `function() {}` to `Function`.

davidchambers:
The commits are not cleanly separated. The third commit fixed some issues with the first commit, meaning that `make lint` fails for the first and second commits. Although I rarely run `git bisect`, I do like to keep a tidy commit history. :)

Avaq:
> The commits are not cleanly separated

That's odd, I ran the tests in a clean working tree after every commit. I probably forgot to lint. I'll re-do it.

Makefile
```diff
@@ -56,6 +56,7 @@ lint:
 	  --global module \
 	  --global require \
 	  --global self \
+	  --rule 'indent: [off]' \
```
Avaq: Are we sure we want to do this? `indent` is quite a significant rule, as it applies to every line of code.

davidchambers: As you know, I'm a big fan of linting. As soon as a rule forces me to do something I don't want to do, though, I'm inclined to disable it. I can police formatting during reviews. :)

Please remove the existing occurrence of `/* eslint-disable indent */` in whichever commit adds the `--rule`.

Avaq: It's disabling `key-spacing` in [the other place](https://github.com/sanctuary-js/sanctuary-type-classes/blob/8ac0f396558081d6e2646ccca7c0fc523966e807/index.js#L986). I don't think it'd be appropriate to remove that in the same commit. If you want I can make a new commit. I would prefer keeping it the way it is though.

davidchambers: `key-spacing` doesn't often cause problems, so disabling that rule selectively seems appropriate.

davidchambers:
Is this pull request ready to merge as far as you're concerned, Aldwin?

Avaq:
Yep!

index.js
```diff
@@ -180,72 +180,89 @@
...
-  }
+  //  prefix :: String
+  var prefix = 'fantasy-land/';
 
-  //  funcPath :: (Array String, a) -> Nullable Function
-  function funcPath(path, x) {
-    return _funcPath(true, path, x);
+  //  funcProp :: (String, a) -> Nullable Function
+  function funcProp(prop, value) {
+    var f = value && value[prop];
```
davidchambers: Please use `value != null` as the guard. I like to avoid type coercion. ;)

This might be clearer:

```javascript
return value != null && typeof value[prop] === 'function' ? value[prop] : null;
```

`s/value/x/g` would allow it to fit on a single line. :)

Avaq: :+1:

index.js
```diff
@@ -180,72 +180,91 @@
...
+  //  funcProp :: (String, a) -> Nullable Function
+  function funcProp(prop, x) {
+    return x != null && typeof x[prop] === 'function' ? x[prop] : null;
   }
 
-  //  funcPath :: (Array String, a) -> Nullable Function
-  function funcPath(path, x) {
-    return _funcPath(true, path, x);
+  //  functionImpl :: (String, String) -> Nullable Function
+  function functionImpl(type, name) {
```
Avaq: You'll notice this function is only used once. I kept it around for symmetry. However, once it's taken into use elsewhere, it'll have to include a check similarly to the one in `methodImpl` to avoid bogus input leading to an exception. Maybe I should just remove it?

index.js
```diff
@@ -180,72 +180,91 @@
...
+    return implementations[type][name];
```
davidchambers: Is this safe? Is it impossible for `implementations[type]` to evaluate to `undefined`?

Avaq: It *is* possible, but only in cases where the code would crash moments later:

With safeguard:

```js
> Z.of(function Foo(){}, 1);
TypeError: Applicative.functions.of(...) is not a function
```

Without safeguard:

```js
> Z.of(function Foo(){}, 1)
TypeError: Cannot read property 'fantasy-land/of' of undefined
```

Avaq: There are no test case such as the one presented, so `if`-path was never even taken.

davidchambers: Thanks for the explanation. Crashing seems okay here. :)

I think the type signature is lying, as this function can—I believe—return `undefined` (which is not a member of `Nullable Function`).

Avaq: This highlights the odd nature of having `getFunction` or `getBoundMethod` return a Nullable. We assign the return `null` to a property which expects to be a function, and then let the code fail later down the line.

I think it would be better to remove the Nullable, and return a default function which throws a descriptive error message.

davidchambers: > I think it would be better to remove the Nullable, and return a default function which throws a descriptive error message

This sounds wonderful! Please do this. :heart:

Avaq: So actually, the fact that this operation is unsafe is preferable to it being safe.

Avaq: All this doesn't count for `methodImpl`, for the simple reason that it's used by `toString`, which performs a null-check on the output.

Avaq: I've convinced myself that `functionImpl` should be in-lined. The next step is to rid ourselves of `methodImpl`. That's in the making.

index.js
```diff
@@ -1122,7 +1141,15 @@
         if (result != null && result !== '[object Object]') return result;
       }
 
-      return call(implPath([xType, 'prototype', 'toString']) || x.toString, x);
+      var f = implementations[xType] != null &&
+              has('toString', implementations[xType].prototype) ?
+              implementations[xType].prototype.toString :
+              typeof x.toString === 'function' ?
+              x.toString :
+              Object.prototype.toString;
```
Avaq: This is here for two reasons:

1. We fix:

```js
> var Bad = function(){}
> Bad['@@type'] = 'NotObject'
> Bad.prototype.toString = 'nonsense'

> Z.toString(new Bad)
TypeError: method.call is not a function
```

Now this code evaluates to `[object Object]`.

2. `methodImpl` no longer needs to check for `hasOwnProperty`. The only culprit was `toString`.

index.js
```diff
@@ -180,72 +180,91 @@
...
+    if (implementations[type] == null) return null;
+    return implementations[type][name] || null;
   }
 
-  //  implPath :: Array String -> Nullable Function
-  function implPath(path) {
-    return _funcPath(false, path, implementations);
+  //  methodImpl :: (String, String) -> Nullable Function
+  function methodImpl(type, name) {
+    if (implementations[type] == null) return null;
```
davidchambers: What happens if `type` is `'toString'` or `'valueOf'`?

Avaq: It throws an exception. This would happen if somebody were to construct a pretty silly argument:

```js
> Z.map(x => x, {constructor: {'@@type': 'toString'}})
TypeError: Cannot read property 'fantasy-land/map' of undefined
```

Avaq: This is no longer a problem, as I reverted the change which introduced it.

Avaq:
I cleaned up the mess:

* Reverted my "fixes" for toString, which might become the subject for another PR.
* Squashed all changes and feedback into two independent commits

index.js
```diff
@@ -180,72 +180,92 @@
...
+    var proto = implementations[type].prototype;
+    return has(name, proto) ? proto[name] : null;
```
Avaq: Oh dear. The problem actually moved here:

```js
> Z.Functor.test({constructor: {'@@type': 'toString'}})
TypeError: Cannot convert undefined or null to object
```

index.js
```diff
@@ -112,6 +107,13 @@
     };
   }
 
+  //  clean :: Object -> Object
+  function clean(x) {
+    var k, o = Object.create(null);
+    for (k in x) o[k] = x[k];
```
Avaq: In my draft, we had the `has` check here, but it turns out we don't need it, as we're passing objects where no properties in the prototype chain are enumerable. I'm not sure if I should add it back as a guard against mutated prototypes, as has been good practice for a long time.

davidchambers: An own-property check here would be reasonable.

Also, I like to pretend we have block-scoped variables, so I would prefer `for (var k in x)` since `k` is only referenced in the loop.

Avaq:
Included `has` check and rebased on top of URL changes.

index.js
```diff
@@ -192,78 +202,97 @@
...
+    return x != null && typeof x[prop] === 'function' ? x[prop] : null;
   }
 
-  //  funcPath :: (Array String, a) -> Nullable Function
-  function funcPath(path, x) {
-    return _funcPath(true, path, x);
+  //  functionImpl :: (String, String) -> Nullable Function
+  function functionImpl(type, name) {
+    return (implementations[type] != null &&
+            implementations[type][name]) || null;
```
davidchambers: I find code clearer when an operator's operands are all on one line or begin at the same column on different lines:

```javascript
foo || bar
```

```javascript
some relatively long expression ||
another relatively long expression
```

_(I apply the same rule to function arguments.)_

My objection to the code above is that `|| null` can be misread as relating to `implementations[type][name]` rather than to the whole parenthesized expression.

This looks ugly, but I prefer it:

```javascript
return (implementations[type] != null &&
        implementations[type][name]) ||
       null;
```

Left to my own devices I might remove the type coercion:

```javascript
return implementations[type] != null &&
       implementations[type][name] != null ?
         implementations[type][name] :
         null;
```

I'm sure the inefficiency of this version would bother you, Aldwin. I'm happy to stick with the coercive version, but please update the indentation. :)

index.js
```diff
@@ -112,6 +112,16 @@
...
+  //  clean :: Object -> Object
+  function clean(x) {
+    var o = Object.create(null);
+    for (var k in x) {
+      /* istanbul ignore else */
+      if (has(k, x)) o[k] = x[k];
+    }
+    return o;
+  }
+
```
davidchambers: Unless you had a reason for inserting this function here, let's preserve alphabetical order. :)

Avaq: Ahh, that's how they are sorted. Of course! I tried to identify some kind of categorization and appended the function to what intuitively felt like a category. ;)
:zap:

index.js
Avaq: My train of thought would be:

1. `functionImpl` is for accessing the `implementations`, it knows everything about this structure
1. all values on `implementations` are guaranteed to be truthy, as we'll only ever assign objects and functions
1. great! no need to build guards against type coercion bugs
1. `return implementations[type] && implementations[type][name] || null`

I can see why you have more contributors than me ;)

:zap:

davidchambers: I find your argument persuasive. I was (and remain) happy with the coercive version. :)

index.js
```diff
@@ -221,78 +228,112 @@
...
+           isPrototype(value) === false &&
+           funcProp(name, value) != null;
+  }
+
+  //  makeFunction :: String -> TypeRep a -> Function
+  function makeFunction(_name) {
+    return function(typeRep) {
+      return getFunction(_name, typeRep);
+    };
+  }
```
davidchambers: `makeFunction` is just `getFunction` curried. Does having access to the uncurried function make a significant difference to performance? If so, could we name this function `getFunctionCurried` or similar?

test/index.js
```diff
@@ -331,6 +331,7 @@ test('Ord', function() {
   eq(Z.Ord.test([]), true);
   eq(Z.Ord.test({}), true);
   eq(Z.Ord.test(Math.abs), false);
+  eq(Z.Ord.test({constructor: {'@@type': 'toString'}}), false);
```
davidchambers: What does this test, exactly?

test/index.js
```diff
@@ -353,6 +355,7 @@ test('Category', function() {
   eq(Z.Category.test([]), false);
   eq(Z.Category.test({}), false);
   eq(Z.Category.test(Math.abs), true);
+  eq(Z.Category.test({constructor: function NotCategory() {}}), false);
```
davidchambers: Would it be appropriate to use [`Useless`][1] here instead?


[1]: https://github.com/sanctuary-js/sanctuary-useless

Avaq:
With the conflicting changes on master, and with #82 showing up, I think I'm not going to move forward with this. Maybe I'll start over at some point in the future. I'd rather put energy in moving forward with #82, so as to make smaller steps.

PR #58, opened 2017-05-23
Update Extend instance for Array

i-am-tom:
Previously, the extend instance was simply `f => [f(this)]`, which is essentially `pure . f`. The common "comonadic" version is introduced in this PR, along with updated tests.

### DIFF
--- a/index.js
+++ b/index.js
@@ -796,7 +796,7 @@
 
   //  Array$prototype$extend :: Array a ~> (Array a -> b) -> Array b
   function Array$prototype$extend(f) {
-    return [f(this)];
+    return this.map(function(_, idx, xs) { return f(xs.slice(idx)); });
   }
 
   //  Arguments$prototype$toString :: Arguments ~> String
@@ -1791,8 +1791,8 @@
   //. built-in types: Array.
   //.
   //. ```javascript
-  //. > extend(xs => xs.length, ['foo', 'bar', 'baz', 'quux'])
-  //. [4]
+  //. > extend(ss => ss.join(''), ['x', 'y', 'z'])
+  //. ['xyz', 'yz', 'z']
   //. ```
   function extend(f, extend_) {
     return Extend.methods.extend(extend_)(f);
--- a/test/index.js
+++ b/test/index.js
@@ -66,6 +66,15 @@ function inc(x) {
   return x + 1;
 }
 
+//  joinWith :: String -> Array String -> String
+function joinWith(s) {
+  eq(arguments.length, joinWith.length);
+  return function joinWith$1(ss) {
+    eq(arguments.length, joinWith$1.length);
+    return ss.join(s);
+  };
+}
+
 //  length :: List a -> Integer
 function length(xs) {
   eq(arguments.length, length.length);
@@ -1054,8 +1063,10 @@ test('extend', function() {
   eq(Z.extend.length, 2);
   eq(Z.extend.name, 'extend');
 
-  eq(Z.extend(length, []), [0]);
-  eq(Z.extend(length, [1, 2, 3]), [3]);
+  eq(Z.extend(joinWith(''), []), []);
+  eq(Z.extend(joinWith(''), ['x']), ['x']);
+  eq(Z.extend(joinWith(''), ['x', 'y']), ['xy', 'y']);
+  eq(Z.extend(joinWith(''), ['x', 'y', 'z']), ['xyz', 'yz', 'z']);
   eq(Z.extend(function(id) { return Z.reduce(add, 1, id); }, Identity(42)), Identity(43));
 });
 


Avaq:
You've forgotten to update the related code examples, which reside in the code-comments in `index.js`, which is why the tests are failing.

i-am-tom:
Aha! Thank you. Two seconds.

index.js
```diff
@@ -786,7 +786,10 @@
 
   //  Array$prototype$extend :: Array a ~> (Array a -> b) -> Array b
   function Array$prototype$extend(f) {
-    return [f(this)];
+    var xs = this;
+    return xs.map(function(_, i) {
+      return f(xs.slice(i));
+    });
```
davidchambers: I find this approach slightly nicer, as it obviates the need for a variable to keep track of `this`:

```javascript
return this.map(function(_, idx, xs) { return f(xs.slice(idx)); });
```

_Note the use of `idx` rather than `i` to match existing code._

index.js
```diff
@@ -1782,7 +1785,7 @@
   //.
   //. ```javascript
   //. > extend(xs => xs.length, ['foo', 'bar', 'baz', 'quux'])
-  //. [4]
+  //. [4, 3, 2, 1]
```
davidchambers: This example is not very informative. How about the following example?

```javascript
> Z.extend(ss => ss.join(''), ['x', 'y', 'z'])
['xyz', 'yz', 'z']
```

test/index.js
```diff
@@ -1017,8 +1017,8 @@ test('extend', function() {
   eq(Z.extend.length, 2);
   eq(Z.extend.name, 'extend');
 
-  eq(Z.extend(length, []), [0]);
-  eq(Z.extend(length, [1, 2, 3]), [3]);
+  eq(Z.extend(length, []), []);
+  eq(Z.extend(length, [1, 2, 3]), [3, 2, 1]);
```
davidchambers: Let's change this to something like `eq(Z.extend(ss => ss.join(''), ['x', 'y', 'z']), ['xyz', 'yz', 'z'])` to assert that the correct array slices are provided.

index.js
davidchambers: @safareli may insist on avoiding all this slicing. :hocho: :watermelon:

If he does we could use this approach:

```javascript
var result = [];
for (var xs = this.slice(); xs.length > 0; xs.shift()) result.push(f(xs));
return result;
```

safareli: @davidchambers this would be definitely better, but client might store the all `xs`es and we would mutate it.
in PS List is Extend and it's linear as it's going from right to left: https://github.com/purescript/purescript-lists/blob/v4.0.1/src/Data/List/Lazy/Types.purs#L161-L170

but we it would make no sense for array so this version is fine (btw there is no extend instance for array in PS)

i-am-tom: @safareli I'm adding an array one in the next day or so - spoken to @garyb about it. It's gonna be preeeetty ugly, but slicing seems to be the most performant way. :/

davidchambers: :zap:

index.js
davidchambers: :zap:

test/index.js
davidchambers: :zap:

i-am-tom:
Oh, thanks, @davidchambers! :)

davidchambers:
@i-am-tom, I've pushed some changes to your branch. Please check that you're happy with them. If you are, please run the following commands to update this pull request:

```console
$ git pull git@github.com:i-am-tom/sanctuary-type-classes.git master
$ git rebase 3042ef049059ad66888f76973bbc1c2606ed1ec4^ --interactive
$ # squash commits
$ git pull --rebase git@github.com:sanctuary-js/sanctuary-type-classes.git master
$ git push git@github.com:i-am-tom/sanctuary-type-classes.git master --force
```

I suggest creating feature branches for future pull requests so your `master` branch can act as a conduit for upstream changes. Assuming `origin` and `upstream` aliases for your fork and the main repository respectively, the workflow would be as follows:

```console
$ git checkout master
$ git pull upstream master
$ git push origin master  # optional
$ git checkout array-extend
$ git rebase master  # rebase `array-extend` on (updated) `master`
$ git push origin array-extend --force
```

The other advantage of this approach is that it avoids `--force` pushing to `master`, a scary habit!

davidchambers:
I'm happy to update this pull request on your behalf if you'd like me to do so, @i-am-tom. I don't mess with other people's commits without consent, though.

i-am-tom:
Hey, @davidchambers, work is sucking ALL of my time right now, so you have my full consent to do as you wish (in fact, feel free to mess with any of my commits in any PRs!) I _will_ get round to this, but it might not be for a couple of days, so I'll totally understand if you'd rather just get it merged in! Sorry about this :(

PR #60, opened 2017-05-26
Add benchmarking

Avaq:
Adds benchmarks inpired by @safareli's [work on Daggy](https://github.com/fantasyland/daggy/tree/master/bench)

I created a sub-package in order to be able to do separate dependency management. This is handy to shave some time off the installation of the parent package (as most won't be running the benchmarks), but most importantly, to allow NPM to install `sanctuary-type-classes` as a dependency of itself.

---

I have a few questions about licensing.

1. @safareli published his work [under the MIT licence](https://github.com/fantasyland/daggy/blob/master/package.json#L12). Since I'm including "a significant portion" of his code in this PR, shouldn't I be including the Daggy legal notice? And if so, the actual license text does not appear to be present in Daggy.
2. Which license should I use in this sub-package? I've currently set it to MIT, so not to publish anything with reduced restrictions over the license of the parent package. But I would like to take this code and use it in other projects, without having to include any legal notices. I therefore would prefer to change the license to WTFPL and include a `LICENCE` file. I'm not sure if that's the right approach, though.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
+/bench/node_modules/
 /coverage/
 /node_modules/
--- a/bench/.npmrc
+++ b/bench/.npmrc
@@ -1,0 +1,2 @@
+package-lock=false
+save=false
--- a/bench/README.md
+++ b/bench/README.md
@@ -1,0 +1,14 @@
+# Benchmarks
+
+## Running
+
+* Install node modules required for the benchmark you wish to run
+* Use `npm run bench -- --help` for options
+
+For example, let's say you like to know how the performance of `toString`
+compares to what it was at version 2.x.x:
+
+```console
+$ (cd bench && npm install sanctuary-type-classes@2.x.x)
+$ npm run bench -- --benchmark old-vs-new --match methods.toString.*
+```
--- a/bench/package.json
+++ b/bench/package.json
@@ -1,0 +1,7 @@
+{
+  "name": "sanctuary-type-classes-benchmarks",
+  "version": "0.0.0",
+  "description": "A package.json for the purpose of being able to install dependencies in here",
+  "license": "MIT",
+  "private": true
+}
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,4 +1,14 @@
 {
   "root": true,
-  "extends": ["./node_modules/sanctuary-style/eslint-es3.json"]
+  "extends": ["./node_modules/sanctuary-style/eslint-es3.json"],
+  "overrides": [
+    {
+      "files": ["bench/**/*.js"],
+      "env": {"node": true},
+      "rules": {
+        "key-spacing": ["off"],
+        "max-len": ["off"]
+      }
+    }
+  ]
 }
--- a/bench/old-vs-new.js
+++ b/bench/old-vs-new.js
@@ -1,0 +1,33 @@
+'use strict';
+
+var benchmark = require('sanctuary-benchmark');
+
+var oldZ = require('sanctuary-type-classes');
+var newZ = require('..');
+
+
+var Identity = require('../test/Identity');
+
+//  inc :: Number -> Number
+function inc(x) {
+  return x + 1;
+}
+
+//  prep :: StrMap Function -> StrMap (Pair (StrMap a) Function)
+function prep(specs) {
+  return newZ.map(function(f) { return [{}, f]; }, specs);
+}
+
+module.exports = benchmark(oldZ, newZ, {leftHeader: 'old', rightHeader: 'new'}, prep({
+  'functions.of.Array':          function(Z) { Z.of(Array, 42); },
+  'functions.of.Identity':       function(Z) { Z.of(Identity, 42); },
+  'methods.equals.Identity':     function(Z) { Z.equals(Identity(0), Identity(0)); },
+  'methods.equals.Object':       function(Z) { Z.equals({x: 0, y: 0}, {y: 0, x: 0}); },
+  'methods.map.Array':           function(Z) { Z.map(inc, [1, 2, 3]); },
+  'methods.map.Identity':        function(Z) { Z.map(inc, Identity(1)); },
+  'methods.toString.Identity':   function(Z) { Z.toString(Identity(0)); },
+  'methods.toString.Object':     function(Z) { Z.toString({x: 0, y: 0}); },
+  'methods.toString.String':     function(Z) { Z.toString('hello'); },
+  'test.Comonad.Identity':       function(Z) { Z.Comonad.test(Identity(0)); },
+  'test.Contravariant.Function': function(Z) { Z.Contravariant.test(Math.abs); }
+}));
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "url": "git://github.com/sanctuary-js/sanctuary-type-classes.git"
   },
   "scripts": {
+    "bench": "sanctuary-benchmark",
     "doctest": "sanctuary-doctest",
     "lint": "sanctuary-lint",
     "release": "sanctuary-release",
@@ -18,6 +19,7 @@
   },
   "devDependencies": {
     "fantasy-land": "3.5.0",
+    "sanctuary-benchmark": "1.0.x",
     "sanctuary-scripts": "1.0.x"
   },
   "files": [
--- a/scripts/lint
+++ b/scripts/lint
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+set -euf -o pipefail
+
+node_modules/.bin/sanctuary-lint
+
+node_modules/.bin/eslint \
+  --report-unused-disable-directives \
+  -- bench/*.js


bench/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "scripts": {
+    "start": "node index.js",
+    "test": "node index.js"
+  },
+  "license": "MIT",
+  "dependencies": {
+    "benchmark": "2.1.x",
+    "cli-table": "0.3.x",
+    "colors": "1.1.x",
+    "sanctuary-type-classes": "5.2.0"
```
Avaq: It would be handy to keep this value programmatically synchronized with the `version` in `../package.json`.

davidchambers:
This is exciting! I'll certainly have review comments, but I'll avoid getting into details until we've answered the licensing questions.

Avaq:
## Example output

latest-screenshot

This is a benchmark of the master branch (7fa5c473b6fbf43f350f02fd67d3c3df7b456287) against `5.2.0`, we can see that functions using `String#toString` have increased performance, presumably thanks to be168b33d645c5e661608178df76114fcef3d7a0.

bench/index.js
```diff
@@ -0,0 +1,25 @@
...
+  'functions.of.Identity':       function(Z) { Z.of(Identity, 42); },
+  'methods.equals.Identity':     function(Z) { Z.equals(Identity(0), Identity(0)); },
+  'methods.equals.Object':       function(Z) { Z.equals({x: 0, y: 0}, {y: 0, x: 0}); },
+  'methods.map.Array':           function(Z) { Z.map(inc, [1, 2, 3]); },
+  'methods.map.Identity':        function(Z) { Z.map(inc, Identity(1)); },
+  'methods.toString.Identity':   function(Z) { Z.toString(Identity(0)); },
+  'methods.toString.Object':     function(Z) { Z.toString({x: 0, y: 0}); },
+  'methods.toString.String':     function(Z) { Z.toString('hello'); },
+  'test.Comonad.Identity':       function(Z) { Z.Comonad.test(Identity(0)); },
+  'test.Contravariant.Function': function(Z) { Z.Contravariant.test(Math.abs); }
```
Avaq: This list of tests is an arbitrarily chosen set which should represent a large amount of the code-paths taken.

Avaq: The idea is that we can add specific tests here as we focus on improving the performance of specific scenarios.

Avaq: If the API surface changes, it is possible to specify a test as:

```js
'changed-surafe': {
  Old: function(){ ... },
  New: function(){ ... }
}
```

davidchambers: Should "surafe" be "surface"? If not, what does it mean?

Avaq: It was a typo, but it doesn't matter what the function key is. It's just the name of the test.

safareli:
I guess we shuold add [LICENSE](https://github.com/fantasyland/fantasy-land/blob/master/LICENSE) to daggy but not sure waht should be done here. I don't quite understand how licensing staff should be handled

Avaq:
Another question that needs an answer is whether [`benchmark.js`](https://github.com/sanctuary-js/sanctuary-type-classes/pull/60/files#diff-8d1f6674ce0f29c46d0f13828161642b) *is* actually a significant portion of Daggy. I've modified the code quite heavily. Would you **want** me to refer to the original Daggy copyright, @safareli?

safareli:
Yes it's almost different, do whatever you want :d

Avaq:
License field updated to WTFPL and rebased on master.

Avaq:
> I'll certainly have review comments -- @davidchambers 

Bring it :D

.gitignore
```diff
@@ -1,2 +1,3 @@
 /coverage/
 /node_modules/
+/bench/node_modules/
```
davidchambers: :abc:

Makefile
```diff
@@ -1,3 +1,4 @@
+BENCH = bench
```
davidchambers: The other constants defined here are executable commands. I'd prefer not to define `BENCH`.

bench/benchmark.js
```diff
@@ -0,0 +1,65 @@
+'use strict';
+
+var Table = require('cli-table');
+var colors = require('colors/safe');
+var benchmark = require('benchmark');
```
davidchambers: Let's alphabetize these by package name. :)

Avaq:
Amended to incorporate feedback

Makefile
```diff
@@ -36,6 +36,11 @@ index.js.tmp: index.js
 	sed -e '/^[/][/]:/ s,\($(SED_WORD_START)[[:alnum:]]*\),<a href="#\1">\1</a>,g' -e 's,^//:,//.,' '$<' >'$@'
 
 
+.PHONY: bench
+bench:
+	node bench
```
davidchambers: This assumes that `npm install` has been run in the __bench__ directory. Shall we add something like `cd bench && $(NPM) install` to the `setup` target?

bench/benchmark.js
```diff
@@ -0,0 +1,65 @@
...
+}
+
+function format(res) {
+  return res.hz.toLocaleString('en', {maximumFractionDigits: 0}) + ' Hz '
+       + '±' + res.stats.rme.toFixed(2) + '%'
+       + '(n ' + String(res.stats.sample.length) + ')';
+}
+
+module.exports = function runBenchmarks(specs) {
+  var table = new Table({head: ['name', 'Old', 'New', 'change']});
```
davidchambers: `Old` and `New` are not particularly descriptive. Would it be possible to show something like `v6.0.0` and `avaq/untangle@abc1234` instead? This way the table would not require context in order to be meaningful, so it could be shared more easily (on Gitter, for example).

bench/benchmark.js
```diff
@@ -0,0 +1,65 @@
+'use strict';
+
+var benchmark = require('benchmark');
+var Table = require('cli-table');
+var colors = require('colors/safe');
```
davidchambers: I'd like to avoid this dependency. Defining `colors.red` and `colors.green` ourselves is trivial:

```javascript
function red(s) {
  return '\u001B[31m' + s + '\u001B[0m';
}

function green(s) {
  return '\u001B[32m' + s + '\u001B[0m';
}
```

The useful thing the dependency does for us is determine whether it's appropriate to use colour, in [__supports-colors.js__][1]. The logic could be simplified and abbreviated to the following:

```javascript
var isColorTerm =
  process.stdout.isTTY &&
  /^screen|^xterm|^vt100|color|ansi|cygwin|linux/i.test(process.env.TERM);
```

Perhaps `isColorTerm` should be determined in __bench/index.js__ and provided to `runBenchmarks` as an argument. We could then change this code:

```diff
-      var colorize = change > +0.1 ? colors.green :
-                     change < -0.1 ? colors.red :
+      var colorize = isColorTerm && change > +0.1 ? green :
+                     isColorTerm && change < -0.1 ? red :
                      identity;
```


[1]: https://github.com/Marak/colors.js/blob/v1.1.2/lib/system/supports-colors.js

bench/benchmark.js
```diff
@@ -0,0 +1,65 @@
...
+    minimumFractionDigits: 1,
+    maximumFractionDigits: 1,
+    useGrouping: false
+  })).slice(-6) + '%';
+}
+
+function format(res) {
+  return res.hz.toLocaleString('en', {maximumFractionDigits: 0}) + ' Hz '
+       + '±' + res.stats.rme.toFixed(2) + '%'
+       + '(n ' + String(res.stats.sample.length) + ')';
```
davidchambers: Did you intentionally omit a space between `%` and `(`? It looks a bit cramped to me.

bench/benchmark.js
```diff
@@ -0,0 +1,65 @@
...
+  return x;
+}
+
+function formatPct(pct) {
+  return ('+' + pct.toLocaleString('en', {
+    minimumIntegerDigits: 3,
+    minimumFractionDigits: 1,
+    maximumFractionDigits: 1,
+    useGrouping: false
+  })).slice(-6) + '%';
```
davidchambers: `.slice(-6)` can truncate the most significant digits, can't it?

bench/benchmark.js
Avaq: That was unintentional. :zap:

bench/benchmark.js
Avaq: That's a really ugly way to slice the `+` off the `+-` case. I should rewrite this function in a more sane way.

bench/benchmark.js
```diff
@@ -0,0 +1,65 @@
...
+                  format(oldRes),
+                  colorize(format(newRes)),
+                  colorize(changePct)]);
+    });
+
+    suite.run();
+
+  });
+
+  console.log(table.toString());
```
davidchambers: For consistency, let's use `process.stdout.write` in all three places. I recall `console.log` being problematic due to its string formatting capabilities. I appreciate the straightforwardness of `process.stdout.write`.

bench/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "sanctuary-type-classes-benchmarks",
+  "version": "0.0.0",
+  "main": "index.js",
```
davidchambers: Let's omit this (as we do in the top-level __package.json__) as it's the default value.

bench/package.json
```diff
@@ -0,0 +1,17 @@
...
+  "private": true,
+  "scripts": {
+    "start": "node index.js",
+    "test": "node index.js"
+  },
```
davidchambers: How do you intend these to be used?

bench/package.json
davidchambers: Good idea! This should do the trick:

```bash
node --eval '
  var fs = require("fs");
  var pkg = require("./bench/package.json");
  pkg.dependencies["sanctuary-type-classes"] = process.env.VERSION;
  fs.writeFileSync("./bench/package.json",
                   JSON.stringify(pkg, null, 2) + "\n",
                   {encoding: "utf8"});
'
```

Surprisingly, the `require("fs")` isn't actually necessary (but let's not rely on this surprising behaviour).

While you're making this change, please manually update the version above to the latest version. :)

bench/index.js
```diff
@@ -0,0 +1,25 @@
...
+
+var Identity = require('../test/Identity');
+
+//  inc :: Number -> Number
+function inc(x) {
+  return x + 1;
+}
+
+benchmark({
+  warmup:                        function()  { Identity(inc(253)); },
```
davidchambers: When aligning text at a particular column I like to use a multiple of two—preferably a multiple of four—because I like even numbers (as does my text editor). :)

We're currently at 33. Let's use 36 or 40 instead. This will also reduce the likelihood of needing to shift the column at some point in the future to accommodate a longer key.

bench/index.js
davidchambers: I'd like to use a different separator in the keys. I find `.` confusing because it makes me think of dot notation. How about using `/` instead? It would then be clear that the keys are just text and not lookup paths.

bench/benchmark.js
davidchambers: You could invoke `toLocaleString` on `Math.abs(pct)` and prefix `'+'` or `'-'` manually.

bench/benchmark.js
```diff
@@ -0,0 +1,65 @@
...
+
+    suite.on('complete', function() {
+      var oldRes = this[0], newRes = this[1];
+      var change = (newRes.hz - oldRes.hz) / oldRes.hz;
+      var changePct = formatPct(change * 100);
+      var colorize = change > +0.1 ? colors.green :
+                     change < -0.1 ? colors.red :
+                     identity;
+
+      console.log(' ...', changePct);
```
Avaq: Going to remove this line

bench/benchmark.js
```diff
@@ -0,0 +1,65 @@
...
+  return res.hz.toLocaleString('en', {maximumFractionDigits: 0}) + ' Hz '
+       + '±' + res.stats.rme.toFixed(2) + '%'
+       + '(n ' + String(res.stats.sample.length) + ')';
+}
+
+module.exports = function runBenchmarks(specs) {
+  var table = new Table({head: ['name', 'Old', 'New', 'change']});
+
+  Object.keys(specs).forEach(function(name, i, keys) {
+    process.stdout.write('Running ' + String(i + 1) + '/' + String(keys.length) + ': ' +  name);
```
Avaq: Going to remove this line

Avaq:
Work halted until sanctuary-js/sanctuary-benchmark#1 is merged.

bench/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "sanctuary-type-classes-benchmarks",
+  "version": "0.0.0",
+  "private": true,
+  "license": "WTFPL",
```
davidchambers: Shall we remove this now?

Avaq: I guess we can change it to MIT, now that this "package" only contains code specific to this project. What do you prefer? npm wants us to provide a license.

davidchambers: I take it the only reason we need this file at all is because we want to install the previously published version of sanctuary-type-classes?

Avaq: Indeed

davidchambers: I believe we could instead add `sanctuary-type-classes@5.2.0` as a dev dependency in the top-level __package.json__.

Avaq: When I tried that the first time, npm told me: "Cannot install sanctuary-type-classes as a dependency of itself"

davidchambers: Perhaps it worked in this case because the versions differ: `5.2.0` versus `6.0.0`. If you upgrade the dependency you may see the error return.

Avaq:
I was able to install `sanctuary-type-classes` as a dependency of itself, but only by *manually* editing `package.json` and then running `npm install`. When I run `npm install -D sanctuary-type-classes`, npm will complain that it refuses to install packages as dependencies of themselves.

scripts/prepublish
```diff
@@ -4,4 +4,12 @@ set -e
...
 make
 sed -i '' "s/var version = '.*';/var version = '$VERSION';/" index.js test/index.js
+node --eval '
+  var fs = require("fs");
+  var pkg = require("./package.json");
+  pkg.devDependencies["sanctuary-type-classes"] = process.env.VERSION;
+  fs.writeFileSync("./package.json",
+                   JSON.stringify(pkg, null, 2) + "\n",
+                   {encoding: "utf8"});
+'
```
davidchambers: This is great! It may not be necessary, but it's probably a good idea to add `package.json` to the `git add` arguments list.

Makefile
```diff
@@ -64,6 +70,11 @@ lint:
 	  -- $(PREDOCTEST)
 	$(ESLINT) \
 	  --env node \
+	  --rule 'key-spacing: [off]' \
+	  --rule 'max-len: [off]' \
+	  -- bench/index.js bench/benchmark.js
```
davidchambers: Do these files still exist?

davidchambers:
> I was able to install `sanctuary-type-classes` as a dependency of itself, but only by *manually* editing `package.json` and then running `npm install`.

In that case, might this work?

```make
setup:
	npm install
	npm install 'sanctuary@$(shell node --print 'require("./package.json").version')'
```

Avaq:
>` npm install 'sanctuary@$(shell node --print 'require("./package.json").version')'`

npm doesn't allow this, we'll have to stick with the prepublish script.

Avaq:
In Fluture I did go with the sub-package approach, where [an additional README](https://github.com/fluture-js/Fluture/pull/196/files#diff-7a43ed35a87fba85b3e188877f40f0b6) guides users to install the version they'd like to test against. This has the following advantages:

1. No fighting against npm
1. No extraneous dependencies installed for users who are not interested in certain benchmarks (if any)
1. Freedom to benchmark against any version, rather than just the latest (I've already used this, because I happen to know that the best performing version of Fluture was 5.x).

You had some ideas about automatically including the version numbers in the table headers. I think it's better to leave that up to the user too. They can provide `--left-header v5.0.1 --right-header HEAD` if they'd like. This would onlu be relevant to benchmarks against another version of yourself. Automation might get in the way once we also include suites to test against the performance of other similar libraries.

Avaq:
Not sure why the tests are failing :point_up:

bench/README.md
```diff
@@ -0,0 +1,16 @@
...
+* Use `npm run bench -- --help` for options
+
+For example, let's say you like to know how the performance of `toString`
+compares to what it was at version 2.x:
+
+```sh
+cd bench
+npm i sanctuary-type-classes@2.x
+cd ..
+npm run bench -- --benchmark old-vs-new --match methods.toString.*
```
Avaq: In case you are curious:

latest-screenshot

.config
```diff
@@ -1,2 +1,3 @@
 repo-owner = sanctuary-js
 repo-name = sanctuary-type-classes
+source-files = index.js bench/*.js
```
davidchambers: I'm nervous about this change. It would mean that files in `bench/*.js` would be considered when generating the readme from code comments. It would be safer to provide `scripts/lint` and have it run `eslint` a second time.

bench/.npmrc
```diff
@@ -0,0 +1,2 @@
+package-lock=false
+save=false
```
davidchambers: What role does `save=false` serve?

bench/README.md
davidchambers: Formatting requests:

  - change `sh` to `console`, and prefix each command with a `$` prompt;
  - use `install` rather than `i`; and
  - use `@2.x.x` rather than `@2.x`.

bench/.npmrc
Avaq: It prevents npm from modifying the `package.json` in this directory.

.config
Avaq: Ah yes. I was also nervous and was going to ask you about it.

bench/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "sanctuary-type-classes-benchmarks",
+  "version": "1.0.0",
```
davidchambers: If this is not going to be released as a package, I think `"0.0.0"` would be more appropriate.

bench/.npmrc
davidchambers: Ah, yes. I see now why this is important.

> ```console
> $ cd bench
> $ npm install sanctuary-type-classes@2.x.x
> ```

Avaq:
Ready to merge, @davidchambers?

PR #61, opened 2017-06-15
derive ‘size’ from ‘reduce’

davidchambers:
See sanctuary-js/sanctuary#414 for prior discussion.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -1026,6 +1026,24 @@ test('reduce', function() {
   eq(Z.reduce(Z.concat, 'x', Cons('a', Cons('b', Cons('c', Nil)))), 'xabc');
 });
 
+test('size', function() {
+  eq(Z.size.length, 1);
+  eq(Z.size.name, 'size');
+
+  eq(Z.size([]), 0);
+  eq(Z.size(['foo']), 1);
+  eq(Z.size(['foo', 'bar']), 2);
+  eq(Z.size(['foo', 'bar', 'baz']), 3);
+  eq(Z.size(Nil), 0);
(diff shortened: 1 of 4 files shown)


safareli:
@davidchambers Tuple es well as These might return unpredictable result. technically you can define size for betraversables too :d  (but we don't have bitraversable in FL)

test/index.js
```diff
@@ -1026,6 +1026,24 @@ test('reduce', function() {
...
+  eq(Z.size(['foo', 'bar']), 2);
+  eq(Z.size(['foo', 'bar', 'baz']), 3);
+  eq(Z.size(Nil), 0);
+  eq(Z.size(Cons('foo', Nil)), 1);
+  eq(Z.size(Cons('foo', Cons('bar', Nil))), 2);
+  eq(Z.size(Cons('foo', Cons('bar', Cons('baz', Nil)))), 3);
+  eq(Z.size(Identity('quux')), 1);
+  eq(Z.size(Nothing), 0);
+  eq(Z.size(Just(0)), 1);
+  eq(Z.size(Tuple('abc', 123)), 1);
```
davidchambers: > Tuple […] might return unpredictable result.

I agree that this is surprising initially, but it seems correct. Haskell agrees:

```haskell
> length ("abc", 123)
1
```

What do you suggest, @safareli?

gabejohnson: It's not unpredictable _if_ the user is familiar with how `reduce` works for their data type. `Foldable` is in the signature after all.

safareli: > reduce works for their data type

Exactly!

PR #62, opened 2017-06-16
derive ‘append’ and ‘prepend’ from ‘concat’ and ‘of’

davidchambers:
`append` and `prepend` are defined in Sanctuary, but based on <https://github.com/sanctuary-js/sanctuary/pull/414#issuecomment-308724784> we'd like to move the definitions here. :)

### DIFF
--- a/test/Maybe.js
+++ b/test/Maybe.js
@@ -13,6 +13,14 @@ function _Maybe(tag, value) {
   this.isNothing = tag === 'Nothing';
   this.isJust = tag === 'Just';
   if (this.isJust) this.value = value;
+
+  if (this.isNothing || Z.Setoid.test(this.value)) {
+    this[FL.equals] = Maybe$prototype$equals;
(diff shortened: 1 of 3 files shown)


test/Maybe.js
```diff
@@ -27,10 +35,16 @@ Maybe[FL.of] = Maybe.Just;
 
 Maybe[FL.zero] = Maybe[FL.empty];
 
-Maybe.prototype[FL.equals] = function(other) {
+function Maybe$prototype$equals(other) {
```
gabejohnson: Why this change?

davidchambers: For consistency with changes made in sanctuary-js/sanctuary#388 which were made for the same reason as sanctuary-js/sanctuary#359. Essentially, `Maybe a` satisfies the requirements of Setoid if `a` satisfies the requirements of Setoid.

PR #66, opened 2017-09-12
derive ‘elem’ from ‘equals’ and ‘reduce’

davidchambers:

### DIFF
--- a/index.js
+++ b/index.js
@@ -1853,6 +1853,42 @@
     return reduce(function(n, _) { return n + 1; }, 0, foldable);
   }
 
+  //# elem :: (Setoid a, Foldable f) => (a, f a) -> Boolean
+  //.
+  //. Takes a value and a structure and returns `true` if the
+  //. value is an element of the structure; `false` otherwise.
+  //.
+  //. This function is derived from [`equals`](#equals) and
+  //. [`reduce`](#reduce).
+  //.
+  //. ```javascript
+  //. > elem('c', ['a', 'b', 'c'])
+  //. true
+  //.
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1853,6 +1853,46 @@
...
+  //. false
+  //.
+  //. > elem(0, Nothing)
+  //. false
+  //. ```
+  function elem(x, foldable) {
+    //  Fast path for arrays.
+    if (Array.isArray(foldable)) {
+      return foldable.some(function(y) { return equals(x, y); });
+    }
```
davidchambers: @Avaq, @safareli, and others: are you in favour of this fast path? Having a fast path for `size` is a clear win, but for `elem` the fast path is no better than the general path in the worst case. Is this special case worthwhile?

safareli: Not sure, but if we had short circuited fold then we would have no need for this.

Avaq: I don't think it's worth it. Let the consumer optimise for specific data types. If I would be using `elem` on an array in a hot path I would probably optimise it out regardless of whether it has this fast path implemented.

kedashoe: > If I would be using elem on a significantly large array in a hot path I would probably optimise it out regardless of whether it has this fast path implemented.

If one is doing UI stuff, then sure, if you notice things slowing down you might do some optimizing of this sort. But I know for server stuff I write I probably wouldn't take the time even if maybe I should. It's only becoming more and more popular to spin up temporary instances to do one job and then die, so every millisecond costs money. I'd vote to leave it in :)

davidchambers: I find your comment persuasive, Aldwin. I removed the special case.

PR #67, opened 2017-09-12
derive ‘reverse’ from ‘concat’, ‘empty’, ‘of’, and ‘reduce’

davidchambers:

### DIFF
--- a/index.js
+++ b/index.js
@@ -1889,6 +1889,29 @@
                   foldable);
   }
 
+  //# reverse :: (Applicative f, Foldable f, Monoid (f a)) => f a -> f a
+  //.
+  //. Reverses the elements of the given structure.
+  //.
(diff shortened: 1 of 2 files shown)


davidchambers:
Thanks for your review, Aldwin.

I'd be very surprised if `Z.reverse([['A', 'a'], ['B', 'b'], ['C', 'c']])` were to evaluate to `[['c', 'C'], ['b', 'B'], ['a', 'A']]`. Similarly, I expect `Z.reverse([[1, 2, 3]])` to evaluate to `[[1, 2, 3]]` rather than `[[3, 2, 1]]`. If one sees the Maybe type as constraining a list's length to [0, 1], it follows that `Z.reverse(S.Just([1, 2, 3]))` should evaluate to `Just([1, 2, 3])`.

PR #68, opened 2017-09-12
derive ‘sort’ and ‘sortBy’ from ‘lte’, ‘concat’, ‘empty’, ‘of’, ‘reduce’

davidchambers:

### DIFF
--- a/index.js
+++ b/index.js
@@ -1912,6 +1912,75 @@
                   foldable);
   }
 
+  //# sort :: (Ord a, Applicative f, Foldable f, Monoid (f a)) => f a -> f a
+  //.
+  //. Performs a [stable sort][] of the elements of the given structure,
+  //. using [`lte`](#lte) for comparisons.
+  //.
+  //. This function is derived from [`lte`](#lte), [`concat`](#concat),
+  //. [`empty`](#empty), [`of`](#of), and [`reduce`](#reduce).
+  //.
+  //. See also [`sortBy`](#sortBy).
+  //.
+  //. ```javascript
+  //. > sort(['foo', 'bar', 'baz'])
+  //. ['bar', 'baz', 'foo']
+  //.
+  //. > sort([Just(2), Nothing, Just(1)])
+  //. [Nothing, Just(1), Just(2)]
+  //.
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -1853,6 +1853,72 @@
...
+  //. [`empty`](#empty), [`of`](#of), and [`reduce`](#reduce).
+  //.
+  //. See also [`sort`](#sort).
+  //.
+  //. ```javascript
+  //. > sortBy(s => s.length, ['red', 'green', 'blue'])
+  //. ['red', 'blue', 'green']
+  //.
+  //. > sortBy(s => -s.split(/e/g).length, ['red', 'green', 'blue'])
+  //. ['green', 'red', 'blue']
```
davidchambers: I don't love this example. It's unclear. I'd like to show [stable sorting][1] in action (`'red'` and `'blue'` are tied in this case as both contain one _e_). Suggestions for clearer examples would be appreciated.


[1]: https://en.wikipedia.org/wiki/Sorting_algorithm#Stability

Avaq: Can't you just add another colour to the example above?

davidchambers: Good idea!

index.js
```diff
@@ -1912,6 +1912,75 @@
...
+      while (lower < upper) {
+        var idx = Math.floor((lower + upper) / 2);
+        if (lte(xs[idx].fx, fx)) lower = idx + 1; else upper = idx;
+      }
+      xs.splice(lower, 0, {x: x, fx: fx});
+      return xs;
+    }, [], foldable);
+
+    var F = foldable.constructor;
+    var result = empty(F);
```
davidchambers: @gabejohnson, @safareli, and @syaiful6, should this be `zero(F)` rather than `empty(F)` in accordance with #37?

safareli: If we have constraint `Monoid (f a)` then it should be empty of monoid.

safareli: not `Alternative` could work so not sure

davidchambers: Thanks for your input, Irakli.

I was confused. I believe `Monoid (f a)` is correct here since we're using `concat` (and `empty`). #37 relates to the `chain`-based `filterM` which is an alternative to the `concat`-based `filter`.

PR #69, opened 2017-09-17
derive ‘takeWhile’ and ‘dropWhile’ from ‘filter’

davidchambers:

Commit message:

> This commit also improves the performance of `filter` for arrays.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1728,6 +1728,8 @@
   //. Cons(1, Cons(3, Nil))
   //. ```
   function filter(pred, m) {
+    //  Fast path for arrays.
+    if (Array.isArray(m)) return m.filter(function(x) { return pred(x); });
     var M = m.constructor;
     return reduce(function(m, x) { return pred(x) ? concat(m, of(M, x)) : m; },
                   empty(M),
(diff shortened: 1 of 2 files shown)


gabejohnson:
Looks good, assuming you don't mind losing the short-circuiting for `takeWhile`.

davidchambers:
I don't mind losing short-circuiting, Gabe. Best-case optimizations are the least important.

index.js
```diff
@@ -1981,6 +1983,56 @@
     return result;
   }
 
+  //# takeWhile :: (Applicative f, Foldable f, Monoid (f a)) => (a -> Boolean, f a) -> f a
+  //.
+  //. Discards the first inner value which does not satisfy the predicate, and
+  //. all subsequent inner values.
+  //.
+  //. This function is derived from [`concat`](#concat), [`empty`](#empty),
+  //. [`of`](#of), and [`reduce`](#reduce).
```
Avaq: Is there a reason you're referring to the lowest level functions, rather than just mentioning "filter"?

gabejohnson: Maybe to explain the class constraints?

davidchambers: > Is there a reason you're referring to the lowest level functions, rather than just mentioning "filter"?

I see three reasons:

  - It is consistent with our descriptions of other derived functions. This is not a compelling reason, of course, as we could easily update these.

  - It assists readers in determining that derived functions are derived from Fantasy Land methods exclusively. Readers familiar with Fantasy Land need not follow internal links.

  - The decision to derive a function from underlying Fantasy Land methods *directly* or *indirectly* should be an implementation detail. Switching from one approach to the other should not necessitate documentation changes.

PR #70, opened 2017-09-28
ignore inherited properties when enumerating a string map

### LINKED ISSUE #65: consider ignoring inherited properties in StrMap implementations
@gabejohnson in <https://github.com/sanctuary-js/sanctuary/pull/435#discussion_r135419232>:

> I don't know that we necessarily _should_ match the semantics of `S.concat` as it works on objects in general. I think we may [inadvertently] cause confusion by not explicitly defining `StrMap`. In my opinion inherited properties should not be copied for a `StrMap`.

This is worth discussing. Here's an example which is not completely contrived:

```javascript
function DigitCounter() {}

DigitCounter.count = ($counter, s) => {
  for (let idx = 0; idx < s.length; idx += 1) {
    const c = s[idx];
    if (/\d/.test(c)) $counter[c] += 1;
  }
};

for (let n = 0; n <= 9; n += 1) {
  DigitCounter.prototype[n] = 0;
}

const $counter = new DigitCounter();
DigitCounter.count($counter, String(Math.PI));

$counter;
// => {'1': 2, '2': 1, '3': 3, '4': 1, '5': 3, '6': 1, '7': 1, '8': 1, '9': 3}

Z.map(x => x, $counter);
// => {'0': 0, '1': 2, '2': 1, '3': 3, '4': 1, '5': 3, '6': 1, '7': 1, '8': 1, '9': 3}
```

Note that mapping `x => x` over `$counter` "promotes" the inherited property to an own property. The question is whether the inherited property should be ignored.

I believe the current behaviour is not just defensible but preferable. My reasoning is as follows:

1.  We currently use `for ... in` which respects all enumerable properties (both own and inherited). It ignores non-enumerable properties such as `toString` and `valueOf`. It's unlikely that people are affected one way or the other by our handling of enumerable inherited properties; I imagine that close to 100% of string maps are plain objects.

2.  Although obscure, the example above shows that respecting enumerable inherited properties can be useful. We should not disallow such code without good reason.

3.  `Z.concat` is one of the library's foundational functions. Performing a `hasOwnProperty` check for each key would affect the performance of several other functions as well as of `Z.concat` itself.

I look forward to understanding your reasons for suggesting that all <del>enumerable</del> <ins>inherited</ins> properties should be ignored, @gabejohnson. You may change my mind. :)

sanctuary-js/sanctuary-def#164 contains related discussion about what constitutes a record field.

davidchambers:
Closes #65

### DIFF
--- a/index.js
+++ b/index.js
@@ -97,6 +97,11 @@
     };
   }
 
+  //  forEachKey :: (StrMap a, StrMap a ~> String -> Undefined) -> Undefined
+  function forEachKey(strMap, f) {
+    Object.keys(strMap).forEach(f, strMap);
+  }
+
   //  has :: (String, Object) -> Boolean
   function has(k, o) {
     return Object.prototype.hasOwnProperty.call(o, k);
@@ -214,7 +219,7 @@
   }
 
   //  functionName :: Function -> String
-  var functionName = 'name' in function f() {} ?
+  var functionName = has('name', function f() {}) ?
     function functionName(f) { return f.name; } :
     /* istanbul ignore next */
     function functionName(f) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -214,7 +219,7 @@
   }
 
   //  functionName :: Function -> String
-  var functionName = 'name' in function f() {} ?
+  var functionName = has('name', function f() {}) ?
```
davidchambers: I made this change so `' in '` no longer appears anywhere in the code.

index.js
```diff
@@ -873,22 +878,25 @@
...
     return result;
   }
 
   //  Object$prototype$ap :: StrMap a ~> StrMap (a -> b) -> StrMap b
   function Object$prototype$ap(other) {
     var result = {};
-    for (var k in this) if (k in other) result[k] = other[k](this[k]);
+    forEachKey(this, function(k) {
+      if (has(k, other)) result[k] = other[k](this[k]);
+    });
```
davidchambers: Note that we're replacing `k in other` with `has(k, other)` here. `Z.ap({}, {toString: 42})` behaves correctly as a result.

PR #77, opened 2018-01-02
extend: Add instance for Function

gabejohnson:
~~I could add the `Comonad` instance as well if anyone's interested~~  😦 

Actually I can't as `extract` can't take a `TypeRep`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -995,6 +995,14 @@
     return function(x) { return f(chain(x))(x); };
   }
 
+  //  Function$prototype$extend :: Semigroup a => (a -> b) ~> ((a -> b) -> c) -> (a -> c)
+  function Function$prototype$extend(f) {
+    var extend = this;
+    return function(x) {
+      return f(function(y) { return extend(concat(x, y)); });
+    };
+  }
+
   //  Function$prototype$contramap :: (b -> c) ~> (a -> b) -> (a -> c)
   function Function$prototype$contramap(f) {
     var contravariant = this;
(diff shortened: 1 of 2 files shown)


test/index.js
```diff
@@ -1261,6 +1261,7 @@ test('extend', function() {
   eq(Z.extend(joinWith(''), ['x', 'y']), ['xy', 'y']);
   eq(Z.extend(joinWith(''), ['x', 'y', 'z']), ['xyz', 'yz', 'z']);
   eq(Z.extend(function(id) { return Z.reduce(add, 1, id); }, Identity(42)), Identity(43));
+  eq(Z.extend(function(f) { return f([1, 2]) + '2'; }, function(xs) { return xs.length; })([3, 4]), '42');
```
davidchambers: Let's replace the second function expression with `length` (defined near the top of the file).

index.js
```diff
@@ -1110,6 +1118,7 @@
         'fantasy-land/promap':      Function$prototype$promap,
         'fantasy-land/ap':          Function$prototype$ap,
         'fantasy-land/chain':       Function$prototype$chain,
+        'fantasy-land/extend':      Function$prototype$extend,
```
davidchambers: Please also make this addition:

```diff
   //. `fantasy-land/extend` implementations are provided for the following
-  //. built-in types: Array.
+  //. built-in types: Array and Function.
```

davidchambers: A doctest would be useful too. :)

test/index.js
davidchambers: Actually, the expression above relies on type coercion. Let's avoid this. I'd also like the expression to demonstrate the order of arguments to `concat`. I suggest:

```javascript
> Z.extend(function(f) { return f([3, 4]); }, Z.reverse)([1, 2])
[4, 3, 2, 1]
```

Further improvements much appreciated. :)

index.js
gabejohnson: Done.

test/index.js
gabejohnson: I can't think of any at the moment.

PR #78, opened 2018-01-03
extend: add duplicate

gabejohnson:
`duplicate :: Extend w => w a -> w (w a)` is the categorical dual of `join :: Chain m => m (m a) -> m a`.

Ordinarily I would open an issue before submitting a PR, but I was messing around with `Extend` for #77. ¯\\\_(ツ)_/¯

### DIFF
--- a/index.js
+++ b/index.js
@@ -2137,6 +2137,29 @@
     return Extend.methods.extend(extend_)(f);
   }
 
+  //# duplicate :: Extend w => w a -> w (w a)
+  //.
+  //. Adds one level of nesting to a comonadic structure.
+  //.
+  //. This function is derived from [`extend`](#extend).
+  //.
+  //. ```javascript
+  //. > duplicate(Identity(1))
+  //. Identity(Identity(1))
+  //.
+  //. > duplicate([1])
+  //. [[1]]
+  //.
+  //. > duplicate([1, 2, 3])
+  //. [[1, 2, 3], [2, 3], [3]]
+  //.
+  //. > duplicate(reverse)([1, 2])([3, 4])
+  //. [4, 3, 2, 1]
+  //. ```
+  function duplicate(extend_) {
+    return extend(identity, extend_);
+  }
+
   //# extract :: Comonad w => w a -> a
   //.
   //. Function wrapper for [`fantasy-land/extract`][].
@@ -2231,6 +2254,7 @@
     traverse: traverse,
     sequence: sequence,
     extend: extend,
+    duplicate: duplicate,
     extract: extract,
     contramap: contramap
   };
--- a/test/index.js
+++ b/test/index.js
@@ -51,9 +51,9 @@ function args() {
   return arguments;
 }
 
-//  duplicate :: a -> Pair a a
-function duplicate(x) {
-  eq(arguments.length, duplicate.length);
+//  double :: a -> Pair a a
+function double(x) {
+  eq(arguments.length, double.length);
   return [x, x];
 }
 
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -2125,6 +2125,26 @@
...
+  //. ```javascript
+  //. > duplicate(Identity(1))
+  //. Identity(Identity(1))
+  //.
+  //. > duplicate([[1, 2, 3]])
+  //. [[[1, 2, 3]]]
+  //.
+  //. > duplicate([1, 2, 3])
+  //. [[1, 2, 3], [2, 3], [3]]
+  //. ```
```
davidchambers: I think the order of the second and third examples should be switched, so the unnested example is presented before the nested one. Do you agree?

test/index.js
```diff
@@ -51,9 +51,9 @@ function args() {
...
 }
 
-//  duplicate :: a -> Pair a a
-function duplicate(x) {
-  eq(arguments.length, duplicate.length);
+//  double :: a -> Pair a a
+function double(x) {
+  eq(arguments.length, double.length);
   return [x, x];
 }
```
davidchambers: Thanks for renaming this function. :thumbsup:

davidchambers:
Please rebase this branch now that #77 has been merged. :)

Is it possible to include a test and/or doctest to show `duplicate`-induced *function* nesting?

index.js
gabejohnson: That was intentional as I thought the third example might be non-intuitive. I'm not married to that order though.

index.js
```diff
@@ -2125,6 +2125,26 @@
     return Extend.methods.extend(extend_)(f);
   }
 
+  //# duplicate :: Extend w => w a -> w (w a)
+  //.
+  //. Adds one level of nesting to a comonadic structure.
```
gabejohnson: @davidchambers I'm not fond of this description. I chose it to mirror the description for `join`, but given the `duplicate([1, 2, 3])` example it could be confusing.

index.js
davidchambers: I found the third example surprising, but perhaps more so due to first having read the second example, which hid the surprising behaviour from me. Emphasizing the behaviour for singleton arrays felt disingenuous, but singleton arrays really do have a special place in Fantasy Land thanks to `of`. This seems natural:

```javascript
> Z.duplicate(Z.of(Identity, 42))
Identity(Identity(42))

> Z.duplicate(Z.of(Array, 42))
[[42]]
```

I've changed my mind. I now prefer your order. :)

index.js
davidchambers: I find writing these descriptions very challenging. I think your description is fine, and one of the roles of doctests is to succinctly demonstrate things which are not easily explained in prose.

index.js
davidchambers: Let's change the second example to this:

```javascript
> duplicate([1])
[[1]]
```

This will create a natural progression from Identity value to singleton array to array with multiple elements.

gabejohnson: Done

index.js
```diff
@@ -2137,6 +2137,29 @@
...
+  //. Identity(Identity(1))
+  //.
+  //. > duplicate([1])
+  //. [[1]]
+  //.
+  //. > duplicate([1, 2, 3])
+  //. [[1, 2, 3], [2, 3], [3]]
+  //.
+  //. > duplicate(reverse)([1, 2])([3, 4])
+  //. [4, 3, 2, 1]
```
davidchambers: This example is mind-bending. :)

PR #81, opened 2018-02-01
bifunctor: add mapLeft

gabejohnson:
Per https://github.com/sanctuary-js/sanctuary/issues/476#issuecomment-358470440

I picked `mapLeft` for the name because that seemed the popular choice in the Gitter room. I'm certainly open to changing it.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1671,6 +1671,18 @@
     return Bifunctor.methods.bimap(bifunctor)(f, g);
   }
 
+  //# mapLeft :: Bifunctor f => (a -> b, f a c) -> f b c
+  //.
+  //. Maps the given function over the left side of a Bifunctor.
+  //.
+  //. ```javascript
+  //. > mapLeft(Math.sqrt, Tuple(64, 9))
+  //. Tuple(8, 9)
+  //. ```
+  function mapLeft(f, bifunctor) {
+    return bimap(f, identity, bifunctor);
+  }
+
   //# promap :: Profunctor p => (a -> b, c -> d, p b c) -> p a d
   //.
   //. Function wrapper for [`fantasy-land/promap`][].
@@ -2277,6 +2289,7 @@
     reject: reject,
     map: map,
     bimap: bimap,
+    mapLeft: mapLeft,
     promap: promap,
     ap: ap,
     lift2: lift2,
--- a/test/index.js
+++ b/test/index.js
@@ -951,6 +951,13 @@ test('bimap', function() {
   eq(Z.bimap(toUpper, inc, Tuple('abc', 123)), Tuple('ABC', 124));
 });
 
+test('mapLeft', function() {
+  eq(Z.mapLeft.length, 2);
+  eq(Z.mapLeft.name, 'mapLeft');
+
+  eq(Z.mapLeft(toUpper, Tuple('abc', 'def')), Tuple('ABC', 'def'));
+});
+
 test('promap', function() {
   eq(Z.promap.length, 3);
   eq(Z.promap.name, 'promap');


index.js
```diff
@@ -1608,6 +1608,18 @@
...
+  //# mapLeft :: Bifunctor f => (a -> b, f a c) -> f b c
+  //.
+  //. Maps the given function over the first type argument of a Bifunctor
+  //.
+  //. ```javascript
+  //. > mapLeft(s => s.toUpperCase(), Tuple('foo', 'bar'))
+  //. Tuple('FOO', 'bar')
+  //. ```
+  function mapLeft(f, bifunctor) {
+    return Bifunctor.methods.bimap(bifunctor)(f, identity);
```
gabejohnson: I chose to avoid the extra function call to `bimap`. If everyone agrees with this, it might be a good idea to do the same with other derived functions.

index.js
```diff
@@ -1608,6 +1608,18 @@
     return Bifunctor.methods.bimap(bifunctor)(f, g);
   }
 
+  //# mapLeft :: Bifunctor f => (a -> b, f a c) -> f b c
+  //.
+  //. Maps the given function over the first type argument of a Bifunctor
```
davidchambers: We usually describe functions at the value level. It's not clear how to describe this function without referencing the type level. Any ideas?

Also, this sentence is missing its full stop. ;)

index.js
```diff
@@ -1608,6 +1608,18 @@
...
+  //.
+  //. ```javascript
+  //. > mapLeft(s => s.toUpperCase(), Tuple('foo', 'bar'))
+  //. Tuple('FOO', 'bar')
```
davidchambers: I'm fond of using `Math.sqrt` in examples because it's short and doesn't require a lambda.

index.js
davidchambers: I'm happy to make this change, but I'd like to do it for all functions at once (either before or after this pull request is merged). I dislike inconsistency, even when it is short-lived.

gabejohnson: I'd prefer to make that change before this is merged then.

davidchambers: :thumbsup:

davidchambers: Since #82 is currently on hold, let's use `bimap` here for now so this pull request can be merged. :)

index.js
gabejohnson: > Any ideas?

None. That's how I ended up referencing "type argument". Fantasy Land sidesteps any mention of values.

gabejohnson: "Maps the given function over the left side of a Bifunctor."
?

davidchambers: > "Maps the given function over the left side of a Bifunctor."

:thumbsup:

index.js
gabejohnson: ⚡️

index.js
gabejohnson: ⚡️

davidchambers:
LGTM. Please rebase this branch, Gabe, to make sure that everything passes the stricter checks we introduced in #84. Merging a passing pull request only to see a red cross appear beside the merge commit never feels good.

gabejohnson:
Done. The tests pass in CI, but I get an error:

```
node_modules/.bin/sanctuary-generate-readme: line 33: declare: -A: invalid option
declare: usage: declare [-afFirtx] [-p] [name[=value] ...]
node_modules/.bin/sanctuary-generate-readme: line 25: node_modules/.bin/transcribe: No such file or directory
```

davidchambers:
What's the output of `bash --version` on your computer, Gabe?

gabejohnson:
I'm running v3.2.57 (Apple provided). Looks like it might be a time for an upgrade 😄

davidchambers:
I think so. According to Wikipedia support for associative arrays was added in version 4.

If you use Homebrew, `brew install bash` should get you a suitable version. :)

PR #82, opened 2018-02-01
Avoid calls to dispatch functions for derived functions

gabejohnson:
Every derived function we define makes at least one call to a dispatching function. If sanctuary-type-classes is supposed to be a foundational library, we should try to avoid as many unnecessary operations as we can.

I've changed calls to dispatching `f` to instead be calls to `Algebra.methods.f` with the exceptions of `lte` and `equals`. Those two functions contain additional operations to handle circular data structure comparisons.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1578,7 +1578,8 @@
   //. Nothing
   //. ```
   function reject(pred, filterable) {
-    return filter(function(x) { return !pred(x); }, filterable);
+    var filter = Filterable.methods.filter;
+    return filter(filterable)(function(x) { return !pred(x); });
   }
 
   //# takeWhile :: Filterable f => (a -> Boolean, f a) -> f a
@@ -1602,7 +1603,8 @@
   //. ```
   function takeWhile(pred, filterable) {
     var take = true;
-    return filter(function(x) { return take = take && pred(x); }, filterable);
+    var filter = Filterable.methods.filter;
+    return filter(filterable)(function(x) { return take = take && pred(x); });
   }
 
   //# dropWhile :: Filterable f => (a -> Boolean, f a) -> f a
@@ -1626,7 +1628,8 @@
   //. ```
   function dropWhile(pred, filterable) {
     var take = false;
-    return filter(function(x) { return take = take || !pred(x); }, filterable);
+    var filter = Filterable.methods.filter;
+    return filter(filterable)(function(x) { return take = take || !pred(x); });
   }
 
   //# map :: Functor f => (a -> b, f a) -> f b
@@ -1730,7 +1733,7 @@
   //. Identity(1000)
   //. ```
   function lift2(f, x, y) {
-    return ap(map(f, x), y);
+    return Apply.methods.ap(y)(Functor.methods.map(x)(f));
   }
 
   //# lift3 :: Apply f => (a -> b -> c -> d, f a, f b, f c) -> f d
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -2056,23 +2060,23 @@
...
+    var result = Monoid.methods.empty(F)();
     for (var idx = 0; idx < rs.length; idx += 1) {
-      result = concat(result, of(F, rs[idx].x));
+      result = Semigroup.methods
+        .concat(result)(Applicative.methods.of(F)(rs[idx].x));
     }
     return result;
   }
 
-  //# takeWhile :: (Applicative f, Foldable f, Monoid (f a)) => (a -> Boolean, f a) -> f a
```
gabejohnson: I took the liberty of changing the signature and description of both `takeWhile` and `dropWhile`. They've been out of date since the introduction of `Filterable`.

davidchambers: Thank you!

index.js
```diff
@@ -2056,23 +2060,23 @@
...
+  //# takeWhile :: (Filterable f) => (a -> Boolean, f a) -> f a
```
davidchambers: We follow Haskell's lead by dropping the parens when there's only one constraint.

index.js
```diff
@@ -1563,7 +1563,8 @@
   //. Nothing
   //. ```
   function reject(pred, filterable) {
-    return filter(function(x) { return !pred(x); }, filterable);
+    return Filterable.methods
+      .filter(filterable)(function(x) { return !pred(x); });
```
davidchambers: I would find this approach to line wrapping slightly more pleasing:

```javascript
var filter = Filterable.methods.filter(filterable);
return filter(function(x) { return !pred(x); });
```

Thoughts?

index.js
gabejohnson: ⚡️

index.js
gabejohnson: It does look nicer and a compiler should optimize that away. Right?

gabejohnson: Perhaps:
```js
var filter = Filterable.methods.filter;
return filter(filterable)(function(x) { return !pred(x); });
```
This way the name matches the operation.

davidchambers: I like it!

gabejohnson: Updated

davidchambers:
This pull request now needs to be rebased, @gabejohnson.

gabejohnson:
@davidchambers I've rebased, but I'm wondering if we should wait for sanctuary-benchmarking integration. It would be nice to see what the improvements (if any) there are.

davidchambers:
> I'm wondering if we should wait for `sanctuary-benchmarking` integration. It would be nice to see what improvements (if any) there are.

Good suggestion, Gabe. Replacing

```javascript
ap(ap(map(f, x), y), z)
```

with

```javascript
Apply.methods.ap(z)(Apply.methods.ap(y)(Functor.methods.map(x)(f)))
```

somewhat diminishes the pedagogical value of the implementations, so confirming that these changes make a noticeable difference before sacrificing readability is a good idea. :)

davidchambers:
You should be able to do some benchmarking now, Gabe, thanks to Aldwin's efforts. :)

davidchambers:
Are you interested in picking this up again, @gabejohnson? If not, we could close this pull request.

PR #88, opened 2018-02-26
fix handling of NaN in Number$prototype$lte to guarantee totality

davidchambers:
Currently, `Z.lte(NaN, 0)` and `Z.lte(0, NaN)` both evaluate `false`, violating the totality law.

I arbitrarily decided to make `NaN` the “smallest” `Number` value.

/cc @CrossEye

### DIFF
--- a/index.js
+++ b/index.js
@@ -668,7 +668,7 @@
   function Number$prototype$lte(other) {
     return typeof this === 'object' ?
       lte(this.valueOf(), other.valueOf()) :
-      isNaN(this) && isNaN(other) || this <= other;
+      isNaN(this) || this <= other;
   }
 
   //  Date$prototype$toString :: Date ~> () -> String
--- a/test/index.js
+++ b/test/index.js
@@ -683,7 +683,7 @@ test('lte', function() {
   eq(Z.lte(Infinity, -Infinity), false);
   eq(Z.lte(-Infinity, Infinity), true);
   eq(Z.lte(-Infinity, -Infinity), true);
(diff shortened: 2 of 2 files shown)


Avaq:
Why do we consider `NaN` to be a number, but `null` to be its own special "Null" type?

Avaq:
Ah I found the answer:

```js
> Object.prototype.toString.call(null)
'[object Null]'
> Object.prototype.toString.call(NaN)
'[object Number]'
```

Avaq:
I'm looking for a reason to exclude `NaN` from having Number privileges.

davidchambers:
> I'm looking for a reason to exclude `NaN` from having Number privileges.

I think this library should remain unopinionated, so should support all `Number` values. Libraries which depend on sanctuary-type-classes are free to be opinionated by excluding problematic values where appropriate. Sanctuary is certainly opinionated (it disallows Number objects, for example).

Avaq:
We might end up being forced to form an opinion either way, because of [`Symbol.toStringTag`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/toStringTag) and how we rely on `Object.prototype.toString` to provide reliably information about built-ins.

davidchambers:
I'll give people a day or two to comment on this pull request before I merge it.

CrossEye:
Well, now that it's committed, I finally manage to look at it.

But it certainly seems fine to me.  Somehow you need to deal with this case, and the native behavior is incoherent.  The decision is arbitrary, but at least it's coherent.

PR #89, opened 2018-02-28
use fantasy-laws to test lawfulness of implementations

### LINKED ISSUE #53: test lawfulness of implementations for built-in types
Dependent on fantasyland/fantasy-laws#1

davidchambers:
Closes #53

This pull request still has a few rough edges, but I don't want to wait any longer to share these changes. I'm especially keen for @Avaq and @safareli to review the Traversable and ChainRec tests respectively.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -1,5 +1,7 @@
 'use strict';
 
+var laws = require('fantasy-laws');
+var jsc = require('jsverify');
 var type = require('sanctuary-type-identifiers');
 
 var Z = require('..');
@@ -28,6 +30,20 @@ function Lazy$of(x) {
   return Z.of(Lazy, x);
 }
 
+//  ListArb :: Arbitrary a -> Arbitrary (List a)
+function ListArb(arb) {
+  return jsc.array(arb).smap(function(xs) {
+    var list = Nil;
+    for (var idx = 0; idx < xs.length; idx += 1) list = Cons(xs[idx], list);
+    return list;
+  });
+}
+
+//  MaybeArb :: Arbitrary a -> Arbitrary (Maybe a)
+function MaybeArb(arb) {
+  return jsc.oneof(jsc.constant(Maybe.Nothing), arb.smap(Maybe.Just));
+}
+
 //  Point :: () -> Point
 function Point() {
   eq(arguments.length, Point.length);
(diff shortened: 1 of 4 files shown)


test/index.js
```diff
@@ -1313,3 +1396,933 @@ test('contramap', function() {
 
   eq(Z.contramap(length, inc)('abc'), 4);
 });
+
+function testLaws(specs) {
+  Object.keys(specs).forEach(function(typeClassName) {
+    suite(typeClassName, function() {
+      var spec = specs[typeClassName];
```
Avaq: I like this approach! I might refactor https://github.com/fluture-js/Fluture/pull/220 to use a similar one!
Also, it looks like if you would do `var module = laws[typeClassName]`, you wouldn't need to repeat yourself.

```diff
-testLaws({Setoid: {module: laws.Setoid, laws: {reflexivity: []}}})
+testLaws({Setoid: {reflexivity: []}})
```

test/index.js
```diff
@@ -1313,3 +1396,933 @@ test('contramap', function() {
...
+
+suite('laws', function() {
+
+  suite('Null', function() {
+    testLaws({
+      Setoid: {
+        module: laws.Setoid,
+        laws: {
+          reflexivity: [
+            jsc.constant(null)
```
Avaq: This works because `fantasy-laws` uses `sanctuary-type-classes` dispatchers, correct? That's a bit confusing, because essentially these tests are testing against the latest version of `sanctuary-type-classes` that is used in `fantasy-laws`. So if you would break something, these tests would not catch it until *after* you release a patch.

test/index.js
Avaq: Ah, I scrolled down to the part where modules require arguments. I understand now.

Avaq: I just scrolled to the part where module require arguments. I understand now.

davidchambers: I'm pleased you like this approach. I spent many hours trying different approaches, on several occasions making the code less repetitive only to change my mind and remove the indirection.

At one point I had this:

```javascript
suite('Boolean', function() {
  var arb = jsc.bool;
  testLaws(Z.equals, {
    Setoid: {
      'Setoid a => Arbitrary a': arb
    },
    Ord: {
      'Ord a => Arbitrary a': arb
    }
  });
});
```

An object provided mappings from these descriptively labelled values to positional arguments. A given arbitrary (e.g. `Setoid a => Arbitrary a`) could be used to test several of a type class's laws, and could be used multiple times when testing certain laws (e.g. associativity).

The approach did not work well, though, for more complex laws. The Traversable laws proved particularly troublesome:

```javascript
Traversable: {
  'Traversable t => Arbitrary (TypeRep t)': jsc.constant(List),
  'Traversable u => Arbitrary (TypeRep u)': jsc.constant(Array),
  'Traversable t => Arbitrary (t a)': arb,
  '(Applicative f, Traversable t, Traversable u) => Arbitrary (f (t a) -> f (u a))': jsc.constant(listToArray),
  '(Traversable t, Traversable u) => Arbitrary (u (t a))': jsc.array(ListArb(jsc.number)),
  'Applicative f => Arbitrary (TypeRep f)': jsc.constant(Maybe),
  'Applicative g => Arbitrary (TypeRep g)': jsc.constant(List),
  '(Applicative f, Applicative g, Traversable t) => Arbitrary (t (f (g a)))': jsc.array(MaybeArb(ListArb(jsc.number)))
}
```

The relationships between the values above are not obvious. A more direct use of fantasy-laws is clearer, and I can live with the repetition.

test/index.js
davidchambers: That's a conundrum. Do you see a solution? Should fantasy-laws take `Z` as an argument? Should there be a peer dependency? Could we ensure correctness by following a certain procedure when making changes?

Avaq: One dirty, and partial, fix is to assign all the implementations to their respective fantasy land identifiers before running the tests, thus ensuring that the old `Z` dispatches to the code we want to test, eg:

```js
String.prototype['fantasy-land/concat'] = String$prototype$concat
```

This wouldn't work for `null` or `undefined`, though.

Avaq: I don't think a peer dependency will work, because I'm not sure whether npm will treat the module at `./` as an installed dependency.

Avaq: We could probably do something like `ln -sf . ./node_modules/fantasy-laws/node_modules/sanctuary-type-classes` before running the tests..

davidchambers: > We could probably do something like `ln -sf . ./node_modules/fantasy-laws/node_modules/sanctuary-type-classes` before running the tests.

Good idea. This works. I have updated the pull request. :)

PR #92, opened 2018-03-19
Improve performance of sortBy

paldepind:
This PR improves the performance of `sortBy` and thus also `sort`. These are the benchmark results I get when running `sort` on an array of 100 numbers.

```
┌───────────────────────┬────────────────────┬──────────────────────┬────────┬───────────┬───┐
│ suite                 │ old                │ new                  │ diff   │ change    │ α │
├───────────────────────┼────────────────────┼──────────────────────┼────────┼───────────┼───┤
│ methods.sort.Array    │ 3 Hz ±1.49% (n 12) │ 637 Hz ±0.30% (n 89) │ 099.1% │ +21464.1% │ ✓ │
├───────────────────────┼────────────────────┼──────────────────────┼────────┼───────────┼───┤
│ methods.sort.Foldable │ 4 Hz ±0.99% (n 14) │ 317 Hz ±0.45% (n 82) │ 097.7% │ +8503.4%  │ ✓ │
└───────────────────────┴────────────────────┴──────────────────────┴────────┴───────────┴───┘
```

In this case sorting an array got 214 times faster and sorting a foldable (using List as the benchmark subject) got 85 times faster :smile: 

Note however that to achieve the performance increase I had to use fewer sanctuary-type-class functions internally in the implementation of `sortBy`. It seems that using sanctuary-type-class functions results in some overhead. I created two benchmarks to try and measure the overhead of the functions used inside `sortBy`.
```
var test = R.range(0, 100);
// ...
'methods.test.reduceMin':     function(Z) { Z.reduce(Z.min, 0, test); },
'methods.test.append':        function(Z) { Z.reduce((a, b) => Z.append(b, a), [], [0, 1, 2, 3, 4]); },
```
Then I plugged in Ramda and sanctuary-type-class in the above benchmarks and got the following results:
```
┌───────────────────────┬──────────────────────────┬────────────────────────────┬────────┬───────────┬───┐
│ suite                 │ sanctuary-type-classes   │ Ramda                      │ diff   │ change    │ α │
├───────────────────────┼──────────────────────────┼────────────────────────────┼────────┼───────────┼───┤
│ methods.test.reduceMin│ 2,073 Hz ±0.52% (n 89)   │ 504,551 Hz ±0.37% (n 93)   │ 099.2% │ +24235.2% │ ✓ │
├───────────────────────┼──────────────────────────┼────────────────────────────┼────────┼───────────┼───┤
│ methods.test.append   │ 235,364 Hz ±1.02% (n 87) │ 1,022,981 Hz ±0.46% (n 91) │ 062.6% │ +334.6%   │ ✓ │
└───────────────────────┴──────────────────────────┴────────────────────────────┴────────┴───────────┴───┘
```
Getting the minimum of an array of 100 numbers seems to be 242 times as slow with sanctuary-type-classes and creating an array of size 5 by prepending is about 3.3 times slower than Ramda. `Z.lte` in particular seems to be be slow and it is still used inside `sortBy` in some cases and in these cases `sortBy` is still slow due to `Z.lte` ending up as the bottleneck.

### DIFF
--- a/index.js
+++ b/index.js
@@ -834,7 +834,9 @@
 
   //  Array$prototype$reduce :: Array a ~> ((b, a) -> b, b) -> b
   function Array$prototype$reduce(f, initial) {
-    return this.reduce(function(acc, x) { return f(acc, x); }, initial);
+    var acc = initial;
+    for (var idx = 0; idx < this.length; idx += 1) acc = f(acc, this[idx]);
+    return acc;
   }
 
   //  Array$prototype$traverse :: Applicative f => Array a ~> (TypeRep f, a -> f b) -> f (Array b)
@@ -2150,21 +2152,31 @@
   //. Cons('red', Cons('blue', Cons('green', Nil)))
   //. ```
   function sortBy(f, foldable) {
-    var rs = reduce(function(xs, x) {
-      var fx = f(x);
-      var lower = 0;
-      var upper = xs.length;
-      while (lower < upper) {
-        var idx = Math.floor((lower + upper) / 2);
-        if (lte(xs[idx].fx, fx)) lower = idx + 1; else upper = idx;
-      }
-      xs.splice(lower, 0, {x: x, fx: fx});
-      return xs;
+    var rs = reduce(function(rs, x) {
+      rs.push({idx: rs.length, x: x, fx: f(x)});
+      return rs;
     }, [], foldable);
+
+    var lte_ = (function(r) {
+      switch (typeof (r && r.fx)) {
+        case 'number':  return function(x, y) { return x <= y || x !== x; };
+        case 'string':  return function(x, y) { return x <= y; };
+        default:        return lte;
+      }
+    }(rs[0]));
+
+    rs.sort(function(a, b) {
+      return lte_(a.fx, b.fx) ? lte_(b.fx, a.fx) ? a.idx - b.idx : -1 : 1;
+    });
+
+    if (Array.isArray(foldable)) {
+      for (var idx = 0; idx < rs.length; idx += 1) rs[idx] = rs[idx].x;
+      return rs;
+    }
 
     var F = foldable.constructor;
     var result = empty(F);
-    for (var idx = 0; idx < rs.length; idx += 1) {
+    for (idx = 0; idx < rs.length; idx += 1) {
       result = concat(result, of(F, rs[idx].x));
     }
     return result;
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
   },
   "devDependencies": {
     "fantasy-land": "3.5.0",
+    "list": "2.0.11",
     "sanctuary-benchmark": "1.0.x",
     "sanctuary-scripts": "1.5.x"
   },
--- a/bench/old-vs-new.js
+++ b/bench/old-vs-new.js
@@ -1,12 +1,14 @@
 'use strict';
 
+var L = require('list/fantasy-land');
 var benchmark = require('sanctuary-benchmark');
 
 var oldZ = require('sanctuary-type-classes');
 var newZ = require('..');
 
-
 var Identity = require('../test/Identity');
+var shuffle = require('../test/shuffle');
+
 
 //  inc :: Number -> Number
 function inc(x) {
@@ -18,6 +20,14 @@ function prep(specs) {
   return newZ.map(function(f) { return [{}, f]; }, specs);
 }
 
+var shuffledArray = (function() {
+  var xs = [];
+  for (var x = 0; x < 5000; x += 1) xs.push(x);
+  shuffle(xs);
+  return xs;
+}());
+var shuffledList = L.fromArray(shuffledArray);
+
 module.exports = benchmark(oldZ, newZ, {leftHeader: 'old', rightHeader: 'new'}, prep({
   'functions.of.Array':          function(Z) { Z.of(Array, 42); },
   'functions.of.Identity':       function(Z) { Z.of(Identity, 42); },
@@ -25,6 +35,8 @@ module.exports = benchmark(oldZ, newZ, {leftHeader: 'old', rightHeader: 'new'}, 
   'methods.equals.Object':       function(Z) { Z.equals({x: 0, y: 0}, {y: 0, x: 0}); },
   'methods.map.Array':           function(Z) { Z.map(inc, [1, 2, 3]); },
   'methods.map.Identity':        function(Z) { Z.map(inc, Identity(1)); },
+  'methods.sort.Array':          function(Z) { Z.sort(shuffledArray); },
+  'methods.sort.List':           function(Z) { Z.sort(shuffledList); },
   'methods.toString.Identity':   function(Z) { Z.toString(Identity(0)); },
   'methods.toString.Object':     function(Z) { Z.toString({x: 0, y: 0}); },
   'methods.toString.String':     function(Z) { Z.toString('hello'); },
--- a/test/index.js
+++ b/test/index.js
@@ -1233,6 +1233,8 @@ test('sort', function() {
     eq(Z.sort(Cons('foo', Nil)), Cons('foo', Nil));
     eq(Z.sort(Cons('foo', Cons('bar', Nil))), Cons('bar', Cons('foo', Nil)));
     eq(Z.sort(Cons('foo', Cons('bar', Cons('baz', Nil)))), Cons('bar', Cons('baz', Cons('foo', Nil))));
+    eq(Z.sort([NaN, 3, NaN, 1, NaN, 2, NaN]), [NaN, NaN, NaN, NaN, 1, 2, 3]);
+    eq(Z.sort([Just(3), Just(1), Just(2)]), [Just(1), Just(2), Just(3)]);
   }
 
   runAssertions();
--- a/test/shuffle.js
+++ b/test/shuffle.js
@@ -1,0 +1,13 @@
+'use strict';
+
+//  https://en.wikipedia.org/wiki/Fisher–Yates_shuffle
+
+//  shuffle :: Array a -> Undefined
+module.exports = function(xs) {
+  for (var i = xs.length - 1; i > 0; i -= 1) {
+    var j = Math.floor(Math.random() * (i + 1));
+    var x = xs[i];
+    xs[i] = xs[j];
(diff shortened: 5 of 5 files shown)


index.js
```diff
@@ -2120,24 +2124,41 @@
...
-      result = concat(result, of(F, rs[idx].x));
     }
-    return result;
+    var rs = reduce(reducer, [], foldable);
+    if (rs.length === 0) {
+      return foldable;
+    }
+    var lteFn =
+      typeof rs[0].prop === 'string' || typeof rs[0].prop === 'number' ?
+        function(a, b) { return a <= b; } :
```
Avaq: This function can be hoisted. May slightly benefit performance.

davidchambers:
Amazing work, Simon!

index.js
```diff
@@ -2120,24 +2124,41 @@
...
-    var F = foldable.constructor;
-    var result = empty(F);
-    for (var idx = 0; idx < rs.length; idx += 1) {
-      result = concat(result, of(F, rs[idx].x));
     }
-    return result;
+    var rs = reduce(reducer, [], foldable);
+    if (rs.length === 0) {
+      return foldable;
+    }
```
davidchambers: Does this special case make an appreciable difference? I would have thought that sorting an empty structure would be an efficient operation.

index.js
```diff
@@ -2120,24 +2124,41 @@
...
-      var upper = xs.length;
-      while (lower < upper) {
-        var idx = Math.floor((lower + upper) / 2);
-        if (lte(xs[idx].fx, fx)) lower = idx + 1; else upper = idx;
-      }
-      xs.splice(lower, 0, {x: x, fx: fx});
+    var foldableIsArray = Array.isArray(foldable);
+    var idx = 0;
+    function reducer(xs, x) {
+      xs.push({idx: idx, x: x, prop: f(x)});
```
davidchambers: Do you consider `prop` to be a better name than `fx` for the `f(x)` property?

index.js
davidchambers: I now see that this is guarding `rs[0].prop` below. I would still like to remove the special case. We could define `lteFn` like so:

```javascript
var type = rs.length > 0 ? typeof rs[0].prop : 'undefined';
var lteFn = type === 'number' || type === 'string' ? ... : lte;
```

Avaq: It's because he's unsafely accessing head in other path.

index.js
```diff
@@ -2120,24 +2124,41 @@
...
+    var lteFn =
+      typeof rs[0].prop === 'string' || typeof rs[0].prop === 'number' ?
+        function(a, b) { return a <= b; } :
+        lte;
+    rs.sort(function(a, b) {
+      return lteFn(a.prop, b.prop) ?
+        lteFn(b.prop, a.prop) ? a.idx - b.idx : -1 :
+        1;
+    });
+    if (foldableIsArray) {
```
davidchambers: Since `foldableIsArray` is only referenced here, let's replace this identifier with `Array.isArray(foldable)`.

index.js
```diff
@@ -2120,24 +2124,41 @@
...
+      // If the original foldable was an array we don't need to convert the
+      // array back to a foldable.
+      for (var j = 0; j < rs.length; j += 1) {
+        rs[j] = rs[j].x;
+      }
+      return rs;
+    } else {
+      var result = empty(foldable.constructor);
+      var of = foldable.constructor['fantasy-land/of'];
```
davidchambers: Is this safe? Could `foldable.constructor` be `Function`?

index.js
```diff
@@ -2120,24 +2124,41 @@
...
+      for (var i = 0; i < rs.length; i += 1) {
+        result = result['fantasy-land/concat'](of(rs[i].x));
```
davidchambers: Is this safe? Are we sure that Array is the only built-in type that can support sorting?

package.json
```diff
@@ -19,6 +19,9 @@
   },
   "devDependencies": {
     "fantasy-land": "3.5.0",
+    "knuth-shuffle": "^1.0.8",
+    "list": "^2.0.7",
+    "ramda": "^0.25.0",
```
davidchambers: Please use the `"1.2.x"` form rather than the `"^1.2.3"` form.

index.js
```diff
@@ -827,7 +827,11 @@
 
   //  Array$prototype$reduce :: Array a ~> ((b, a) -> b, b) -> b
   function Array$prototype$reduce(f, initial) {
-    return this.reduce(function(acc, x) { return f(acc, x); }, initial);
+    var acc = initial;
+    for (var i = 0; i < this.length; i += 1) {
+      acc = f(acc, this[i]);
+    }
+    return acc;
   }
```
davidchambers: Does this change make an appreciable difference?

index.js
paldepind: I'm indifferent. For `prop` I didn't actually look at the old name. The thought behind `prop` is that `f` returns some property of `x` (not property as in a JS property of an object but in the more general sense of the word). But I'd have no problem in switching it back.

index.js
paldepind: Good idea. I was using it in several places earlier but that went away.

index.js
davidchambers: I prefer `fx`. :)

index.js
paldepind: I've measured the difference in the [benchmark suite for List](https://funkia.github.io/list/benchmarks/). I'd say the difference is quite significant.

screenshot_20180319_142156

index.js
paldepind: The fast-path for numbers and strings is very important for performance.

I think we should consider whether it's worth the code complexity and performance to handle `NaN`. For instance, `Array$prototype$lte` doesn't handle sparse arrays because sparse arrays can be considered to not be a part of the array type we define. In the same way `Number$prototype$lte` could not handle `NaN` as `NaN` could be considered not part of the number type we define.

paldepind:
@Avaq Props on the benchmark system! The output in particular is really nice :+1:

index.js
davidchambers: Significant indeed!

Now that I know we'll keep the code I'll share my formatting request. I'd like to see this:

```javascript
for (var idx = 0; idx < this.length; idx += 1) acc = f(acc, this[idx]);
```

index.js
davidchambers: I think it would be best for these changes to be refactoring only. In a subsequent pull request you could make the case for not supporting `NaN` by showing the performance difference.

davidchambers:
@paldepind, I'm happy to push tweaks to your branch. I'd love to merge this pull request soon. :)

paldepind:
@davidchambers You're very welcome to that David 👍

index.js
davidchambers: :zap:

index.js
davidchambers: :zap:

index.js
davidchambers: :zap:

index.js
davidchambers: :zap:

index.js
davidchambers: :zap:

paldepind:
Looks good to me @davidchambers.

index.js
```diff
@@ -2132,9 +2132,7 @@
       return xs;
     }
     var rs = reduce(reducer, [], foldable);
-    if (rs.length === 0) return foldable;
-
-    var type = typeof rs[0].fx;
```
paldepind: What is the benefit of this? Replacing one special case with another that doesn't terminate as early?

index.js
```diff
@@ -2140,18 +2140,16 @@
...
-      }
-      return result;
     }
+
+    var F = foldable.constructor;
+    var result = empty(F);
+    for (idx = 0; idx < rs.length; idx += 1) {
+      result = concat(result, of(F, rs[idx].x));
+    }
+    return result;
```
paldepind: This does seem to affect performance. On my computer, the Foldable benchmark went from 308 Hz to 233 Hz. That is, the previous is about 33% faster.

index.js
davidchambers: I prefer `typeof (rs[0] && rs[0].fx)` because the unsafe expression (`rs[0].fx`) and its guard appear very close together. My preference is not strong, though, so feel free to reinstate the `rs.length === 0` check. :)

index.js
davidchambers: I'm concerned about correctness. Representatives of built-in types do not have `fantasy-land/`-prefixed members, so replacing `concat(x, y)` with `x['fantasy-land/concat'](y)` is not necessarily valid. Can we be sure that `result` will have a `fantasy-land/concat` member and that `F` will have a `fantasy-land/of` member?

> `fantasy-land/concat` implementations are provided for the following built-in types: String, Array, and Object.

> `fantasy-land/of` implementations are provided for the following built-in types: Array and Function.

paldepind: Are any of the build-in types, except for Array, both a foldable and a monoid?

gabejohnson: As a `StrMap`, `Object` also implements `Foldable` and `Monoid`.

paldepind: @gabejohnson Sorry, I forgot to mention that it has to be an Applicative as well.

gabejohnson: Ahhh...I missed the `of`. Yeah, `Array` is the only one that fits the bill AFAIK.

davidchambers: If we reinstate the `['fantasy-land/concat']` and `['fantasy-land/of']` optimizations, let's add a comment explaining why these optimizations are safe.

davidchambers:
:information_source: @paldepind, I rebased this branch.

package.json
davidchambers: :zap:

package.json
```diff
@@ -19,6 +19,7 @@
   },
   "devDependencies": {
     "fantasy-land": "3.5.0",
+    "list": "2.0.7",
```
davidchambers: @Avaq, should this be listed here or in __bench/package.json__?

paldepind:
LGTM :tada:

package.json
Avaq: It's used as a library like any other (just so happens to be used in benchmarks), so it should be listed here. The `bench/package.json` file only exists to allow for the installation of any version of choice, of the package you're benchmarking against.

PR #94, opened 2018-03-27
assert that ‘sort’ and ‘sortBy’ do not rely on Array#sort being stable

davidchambers:
While working on #92 I realized I could replace `idx: rs.length` with `idx: 0` and all the tests would still pass. This is because `Array#sort` is stable in my environment, as it is in most environments. The language specification does not guarantee this, though, so we should not assume it.

This pull request introduces a function, `withUnstableArraySort`, which monkey patches `Array#sort`, evaluates a thunk, then restores `Array#sort` to its prior value. Any change to `Z.sort` or `Z.sortBy` which assumes `Array#sort` to be stable will pass tests in the default environment but will fail some tests in the modified environment.

/cc @paldepind

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -11,6 +11,7 @@ var Maybe = require('./Maybe');
 var Sum = require('./Sum');
 var Tuple = require('./Tuple');
 var eq = require('./eq');
+var withUnstableArraySort = require('./quicksort').withUnstableArraySort;
 
 
 var Nil = List.Nil;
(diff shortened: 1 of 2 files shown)


paldepind:
Sorry, I'm late :sob: 

> This is because Array#sort is stable in my environment, as it is in most environments.

`Array#sort` is unstable in V8 from what I can find.

In any case, I think this PR is a very good idea to make the sort predictably unstable :+1:

PR #95, opened 2018-04-06
Derive flip from map

### LINKED ISSUE #63: Derive flip from map
```js
const flip = (f, x) => map(T(x), f)

flip(a => b => a - b, 1)(5) //> 4
flip({a: x => x - 1, b: x => x + 1}, 1) //> {a: 0, b: 2}
```

See https://hackage.haskell.org/package/functors-0.1/docs/Data-Functor-Syntax.html#v:flip

Avaq:
Closes #63

### DIFF
--- a/index.js
+++ b/index.js
@@ -139,6 +139,13 @@
     return typeof x === typeof y && type(x) === type(y);
   }
 
+  //  thrush :: a -> (a -> b) -> b
+  function thrush(x) {
+    return function(f) {
+      return f(x);
+    };
+  }
+
   //  type Iteration a = { value :: a, done :: Boolean }
 
   //  iterationNext :: a -> Iteration a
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1658,6 +1665,31 @@
...
   }
 
+  //# flip :: Functor f => f (a -> b) -> a -> f b
+  //.
+  //. Apply a function contained in a Functor to a value.
+  //.
+  //. This function is derived from [`map`](#map).
+  //.
+  //. ```javascript
+  //. > function div(x) { return function(y) { return x / y }; }
```
Avaq: Not sure what a good example would be, especially considering that the API is a little weird without currying. I imagine Sanctuary would be able to provide a much nicer example for `flip` on Function.

index.js
```diff
@@ -2287,6 +2319,7 @@
     filter: filter,
     reject: reject,
     map: map,
+    flip: flip,
```
Avaq: Not sure where to place the function. It seemed you have been placing derived functions underneath their underlying functions, so I inserted underneath map.

index.js
davidchambers: We're allowed to use arrow functions in doctests. How about the following?

```javascript
> flip(x => y => x + y, '!')('foo')
'foo!'
```

I prefer this example to the `div` example because `div` should really take the divisor as its first argument.

index.js
```diff
@@ -1658,6 +1665,31 @@
     return Functor.methods.map(functor)(f);
   }
 
+  //# flip :: Functor f => f (a -> b) -> a -> f b
+  //.
+  //. Apply a function contained in a Functor to a value.
```
davidchambers: This sentence should begin “Applies…” rather than “Apply…” for consistency with other descriptions. Also, “a function” is misleading as there could be any number of functions. I suggest:

_Returns the result of mapping over the given functions, applying each to the given value._

index.js
davidchambers: Perfect! The order here should match the order in which the functions are defined. :)

index.js
Avaq: We could omit *Returns the result of* from any such description, and in this case simplify to:

*Maps over the given functions, applying each to the given value.*

The fact that the results of that is returned should be clear. What do you think?

davidchambers: Sounds good to me. :)

Avaq:
Amended.

davidchambers:
Thanks. Could you also rebase?

Avaq:
Done.

PR #102, opened 2018-05-05
Change Array$prototype$chain to use for loop instead of forEach

edahlseng:
This pull request modifies Array$prototype$chain to use a for loop instead of forEach, as discussed in https://github.com/sanctuary-js/sanctuary/issues/527.

### DIFF
--- a/index.js
+++ b/index.js
@@ -825,7 +825,9 @@
   //  Array$prototype$chain :: Array a ~> (a -> Array b) -> Array b
   function Array$prototype$chain(f) {
     var result = [];
-    this.forEach(function(x) { Array.prototype.push.apply(result, f(x)); });
+    this.forEach(function(x) {
+        f(x).forEach(function(y) { result.push(y); });
+    });
     return result;
   }
 


edahlseng:
Hi @davidchambers, thanks for that, I'll check out those errors! Would you prefer that the changes are fixed up into the existing commit, or done as a separate commit?

davidchambers:
> Would you prefer that the changes are fixed up into the existing commit, or done as a separate commit?

I don't mind. If you make multiple commits I'll ask you to squash them once the pull request is in good shape. If you decide to `git commit --amend` please let me know when you push changes, as GitHub does not notify me of `--force` pushes.

edahlseng:
@davidchambers just pushed an update. Sorry for the incorrect implementation the first time; I should have actually read the type signature!

Interestingly enough, it looks like `forEach` was not the cause for exceeding the stack size, but rather using `Array.prototype.apply` (I'm not sure why that would case the error; any ideas?). The changes that I've just pushed revert back to using `forEach`, but instead of `Array.prototype.apply` there's `result.push` in a nested `forEach`. It feels a bit gross to me, but I haven't found any cleaner solutions that solve the problem yet.

davidchambers:
> it looks like `forEach` was not the cause for exceeding the stack size, but rather using `Array.prototype.apply`

Nice detective work. :)

Could you add a test case to prove that the change is effective and to prevent regression?

index.js
```diff
@@ -825,7 +825,9 @@
   //  Array$prototype$chain :: Array a ~> (a -> Array b) -> Array b
   function Array$prototype$chain(f) {
     var result = [];
-    this.forEach(function(x) { Array.prototype.push.apply(result, f(x)); });
+    this.forEach(function(x) {
+        f(x).forEach(function(y) { result.push(y); });
+    });
```
davidchambers: We favour indentation of two spaces.

In this case wrapping the body of the innermost lambda would make the nesting most clear:

```javascript
this.forEach(function(x) {
  f(x).forEach(function(y) {
    result.push(y);
  });
});
```

davidchambers:
The title of this pull request is now out of date. You could update it. It can be advantageous for the title to describe the problem the changes address rather than exactly how they do so, as the specifics may change based on review feedback.

davidchambers:
Superseded by #103

PR #103, opened 2018-05-07
remove uses of Function#apply to accommodate large arrays

davidchambers:
Supersedes #102

The new implementation of `Array$chainRec` is longer but arguably simpler and more efficient.

<table>
  <tr>
    <td colspan="2"><b>Before</b></td>
  </tr>
  <tr>
    <td>Data structure</td>
    <td>array</td>
  </tr>
  <tr>
    <td>Loop condition</td>
    <td><code>$todo.length > 0</code></td>
  </tr>
  <tr>
    <td>Advancement</td>
    <td><code>$todo.shift()</code></td>
  </tr>
  <tr>
    <td>Accumulation</td>
    <td><code>$more.push(step.value)</code></td>
  </tr>
  <tr>
    <td>Transference</td>
    <td><code>Array.prototype.unshift.apply($todo, $more)</code> :warning:</td>
  </tr>
  <tr>
    <td colspan="2"><b>After</b></td>
  </tr>
  <tr>
    <td>Data structure</td>
    <td>linked list</td>
  </tr>
  <tr>
    <td>Loop condition</td>
    <td><code>xs !== nil</code></td>
  </tr>
  <tr>
    <td>Advancement</td>
    <td><code>xs = xs.tail</code></td>
  </tr>
  <tr>
    <td>Accumulation</td>
    <td><code>sx = {head: step.value, tail: sx}</code></td>
  </tr>
  <tr>
    <td>Transference</td>
    <td><code>while (sx !== nil) { xs = {head: sx.head, tail: xs}; sx = sx.tail; }</code></td>
  </tr>
</table>

I'd appreciate your :eyes: on this pull request, Irakli. :)

/cc @edahlseng

### DIFF
--- a/bench/old-vs-new.js
+++ b/bench/old-vs-new.js
@@ -10,6 +10,16 @@ var Identity = require('../test/Identity');
 var shuffle = require('../test/shuffle');
 
 
+//  chainRecArrayNumber :: (Number -> c, Number -> c, Number) -> Array c
+function chainRecArrayNumber(next, done, x) {
+  return [done(x), (x <= 1 ? done : next)(x - 1)];
+}
+
+//  double :: a -> Array2 a a
+function double(x) {
+  return [x, x];
+}
+
 //  inc :: Number -> Number
 function inc(x) {
   return x + 1;
@@ -29,8 +39,10 @@ var shuffledArray = (function() {
 var shuffledList = L.fromArray(shuffledArray);
 
 module.exports = benchmark(oldZ, newZ, {leftHeader: 'old', rightHeader: 'new'}, prep({
+  'functions.chainRec.Array':    function(Z) { Z.chainRec(Array, chainRecArrayNumber, 100); },
   'functions.of.Array':          function(Z) { Z.of(Array, 42); },
   'functions.of.Identity':       function(Z) { Z.of(Identity, 42); },
+  'methods.chain.Array':         function(Z) { Z.chain(double, shuffledArray); },
   'methods.equals.Identity':     function(Z) { Z.equals(Identity(0), Identity(0)); },
   'methods.equals.Object':       function(Z) { Z.equals({x: 0, y: 0}, {y: 0, x: 0}); },
   'methods.map.Array':           function(Z) { Z.map(inc, [1, 2, 3]); },
--- a/index.js
+++ b/index.js
@@ -747,17 +747,27 @@
 
   //  Array$chainRec :: ((a -> c, b -> c, a) -> Array c, a) -> Array b
   function Array$chainRec(f, x) {
-    var $todo = [x];
-    var $done = [];
-    while ($todo.length > 0) {
-      var xs = f(iterationNext, iterationDone, $todo.shift());
-      var $more = [];
-      for (var idx = 0; idx < xs.length; idx += 1) {
-        (xs[idx].done ? $done : $more).push(xs[idx].value);
+    var result = [];
+    var nil = {};
+    var todo = {head: x, tail: nil};
+    while (todo !== nil) {
+      var more = nil;
+      var steps = f(iterationNext, iterationDone, todo.head);
+      for (var idx = 0; idx < steps.length; idx += 1) {
+        var step = steps[idx];
+        if (step.done) {
+          result.push(step.value);
+        } else {
+          more = {head: step.value, tail: more};
+        }
       }
-      Array.prototype.unshift.apply($todo, $more);
+      todo = todo.tail;
+      while (more !== nil) {
+        todo = {head: more.head, tail: todo};
+        more = more.tail;
+      }
     }
-    return $done;
+    return result;
   }
 
   //  Array$zero :: () -> Array a
@@ -825,7 +835,11 @@
   //  Array$prototype$chain :: Array a ~> (a -> Array b) -> Array b
   function Array$prototype$chain(f) {
     var result = [];
-    this.forEach(function(x) { Array.prototype.push.apply(result, f(x)); });
+    for (var idx = 0; idx < this.length; idx += 1) {
+      for (var idx2 = 0, xs = f(this[idx]); idx2 < xs.length; idx2 += 1) {
+        result.push(xs[idx2]);
+      }
+    }
     return result;
   }
 
--- a/test/index.js
+++ b/test/index.js
@@ -1177,6 +1177,7 @@ test('chain', function() {
   eq(Z.chain(repeat, abs)(-3), [-3, -3, -3]);
   eq(Z.chain(identInc, Identity(42)), Identity(43));
   eq(Z.chain(identity, Identity(Identity(0))), Identity(0));
+  eq(Z.chain(identity, [new Array(1e6).join('x').split('')]).length, 999999);
 });
 
 test('join', function() {
@@ -1217,6 +1218,8 @@ test('chainRec', function() {
   }
 
   eq(Z.chainRec(Function, stepper, 0)({step: 2, inc: 100}), 30100);
+
+  eq(Z.chainRec(Array, function(next, done, n) { return n === 0 ? [done(0)] : Z.map(next, repeat(n)(0)); }, 1e6).length, 1e6);
 });
 
 test('alt', function() {


Avaq:
:eyes:

bench/old-vs-new.js
```diff
@@ -9,6 +9,15 @@ var newZ = require('..');
 var Identity = require('../test/Identity');
 var shuffle = require('../test/shuffle');
 
+//  chainRec :: (Number -⁠> c, Number -⁠> c, Number) -⁠> Array c, Number) -⁠> Array b
+function chainRecArrayNumber(next, done, x) {
+  return [done(x), (x <= 1 ? done : next)(x - 1)];
+}
+
+//  duplicate :: a -> Pair a a
```
davidchambers: Let's use `Array2` in place of `Pair`.

bench/old-vs-new.js
```diff
@@ -9,6 +9,15 @@ var newZ = require('..');
 var Identity = require('../test/Identity');
 var shuffle = require('../test/shuffle');
 
+//  chainRec :: (Number -⁠> c, Number -⁠> c, Number) -⁠> Array c, Number) -⁠> Array b
```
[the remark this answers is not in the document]
davidchambers: Please remove the WORD JOINER characters from the type signature. I like to stick to ASCII in source code. :)

bench/old-vs-new.js
Avaq: I copied the style from the other signatures in this file. In that case let's change all.

bench/old-vs-new.js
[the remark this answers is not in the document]
Avaq: Oh darn. I copied from the Sanctuary website. :P

index.js
```diff
@@ -747,17 +747,25 @@
...
+        } else {
+          sx = {head: step.value, tail: sx};
+        }
+      });
+      xs = xs.tail;
+      while (sx !== nil) {
+        xs = {head: sx.head, tail: xs};
+        sx = sx.tail;
       }
-      Array.prototype.unshift.apply($todo, $more);
```
syaiful6: if `Function.apply` is the root cause of stack size, to support large array we maybe just need to check the limit of size arguments of function, slice it and apply it in loop, and according to https://stackoverflow.com/questions/22747068/is-there-a-max-number-of-arguments-javascript-functions-can-accept/22747272#22747272 the lowest limit is Chrome, with `0x10000` args.

but i am not sure how to do it though..

davidchambers: That could work, but I'd prefer to stick with the linked list implementation (assuming it performs reasonably) as I find it easier to understand.

bench/old-vs-new.js
davidchambers: :zap:

bench/old-vs-new.js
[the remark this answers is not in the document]
davidchambers: :zap::zap:

index.js
```diff
@@ -747,17 +747,25 @@
...
+      f(iterationNext, iterationDone, xs.head).forEach(function(step) {
+        if (step.done) {
+          result.push(step.value);
+        } else {
+          sx = {head: step.value, tail: sx};
+        }
+      });
+      xs = xs.tail;
+      while (sx !== nil) {
+        xs = {head: sx.head, tail: xs};
```
safareli: The $more was there so during iterations order was "respected". ie result of each "step" was prepended to $todo - conceptually like this `$more <> $todo`. this version using list conseptually does this `flip $more <> $todo`, if `flip` is fine, and we don't want to "respect" order then current version could be changed so apply is not needed:

```diff
//  Array$chainRec :: ((a -> c, b -> c, a) -> Array c, a) -> Array b
function Array$chainRec(f, x) {
  var $todo = [x];
  var $done = [];
  while ($todo.length > 0) {
    var xs = f(iterationNext, iterationDone, $todo.shift());
-    var $more = [];
    for (var idx = 0; idx < xs.length; idx += 1) {
-      (xs[idx].done ? $done : $more).push(xs[idx].value);
+      (xs[idx].done ? $done : $todo).push(xs[idx].value);
    }
-    Array.prototype.unshift.apply($todo, $more);
  }
  return $done;
}```


but I think because of the laws  it's not allowed to reorder staff so we have to respect the order.

Tests are passing because the input to the laws i guess contains same values so order is not checked, if in tests we use ordered values then it would be more clear why we should respect the order.

for example this will pass in all versions as the flip is used when there is `next` returned during iteration

```js
Array.chainRec((next,done, x) => {
  return Array.from({length: x}, (_, k) => done(k+1)); 
}, [1,2,3,4,5,6])
```
should evaluate to `[1,1,2,1,2,3,1,2,3,4,1,2,3,4,5,1,2,3,4,5,6]`

but if stepping function returned next then current version will be broken i think, test using something like this:

```
Array.chainRec((next,done, x) => {
  return Array.from({length: x.v}, (_, k) => x.d ? done(k+1) : next({v:k+1, d:true })); 
}, [{v:2, d:false }
   ,{v:3, d:false }
   ]).join('')
```
I think this should evaluate to `112112123` if implementation is correct

safareli: also old names more todo done are better i think than xs ys ...

bench/old-vs-new.js
```diff
@@ -10,12 +10,22 @@ var Identity = require('../test/Identity');
 var shuffle = require('../test/shuffle');
 
 
+//  chainRecArrayNumber :: (Number -> c, Number -> c, Number -> Array c, Number) -> Array Number
```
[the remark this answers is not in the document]
davidchambers: Thanks! I had been staring at this signature thinking it didn't seem quite right. Updated. :)

index.js
davidchambers: > old names more todo done are better i think than xs ys

You're right. I'll revert to the descriptive names. I named the linked lists `xs` and `sx` to hint that the latter is in reverse order, but this is too cute by half.

> if stepping function returned next then current version will be broken i think

I believe—though I'm not certain—that the new algorithm respects order. Whereas previously we used `Array#unshift` to add the elements of `$more` at the beginning of `$todo`, we now prepend the elements of `sx` (which are in reverse order having been consed) to `xs`. By consing twice we preserve the original order.

Your example required a bit of tweaking, but I believe this is what you wanted to test:

```javascript
> Z.chainRec (
  Array,
  (next, done, x) =>
    Array.from ({length: x.v},
                (_, k) => x.d ? done (k + 1) : next ({v: k + 1, d: true})),
  {v: 5, d: false}
)
[1, 1, 2, 1, 2, 3, 1, 2, 3, 4, 1, 2, 3, 4, 5]
```

Please give the code one more look. Now that I've explained the double consing you may be convinced that order is preserved. If not, let's add a few more test cases to be sure. :)

index.js
```diff
@@ -747,17 +747,25 @@
...
-      var xs = f(iterationNext, iterationDone, $todo.shift());
-      var $more = [];
-      for (var idx = 0; idx < xs.length; idx += 1) {
-        (xs[idx].done ? $done : $more).push(xs[idx].value);
+    var result = [];
+    var nil = {};
+    var todo = {head: x, tail: nil};
+    while (todo !== nil) {
+      var more = nil;
+      f(iterationNext, iterationDone, todo.head).forEach(function(step) {
```
Bradcomp: Why switch to a `forEach` here? 

As much as I _hate_ for loops in application code, in a library I _think_ it's faster than a `forEach`

davidchambers: Now that Aldwin has added benchmarks (:clap:) we can see. I'll compare the two options.

davidchambers: Using a `for` loop here is significantly faster on my computer. Thanks for the suggestion, Brad!

davidchambers:
Performance update:

    ┌──────────────────────────┬─────────────────────────┬──────────────────────────┬────────┬─────────┬───┐
    │ suite                    │ master                  │ davidchambers/no-apply   │ diff   │ change  │ α │
    ├──────────────────────────┼─────────────────────────┼──────────────────────────┼────────┼─────────┼───┤
    │ functions.chainRec.Array │ 83,749 Hz ±0.89% (n 89) │ 272,724 Hz ±0.94% (n 84) │ 053.0% │ +225.6% │ ✓ │
    ├──────────────────────────┼─────────────────────────┼──────────────────────────┼────────┼─────────┼───┤
    │ methods.chain.Array      │ 2,886 Hz ±0.63% (n 88)  │ 12,114 Hz ±0.47% (n 90)  │ 061.5% │ +319.8% │ ✓ │
    └──────────────────────────┴─────────────────────────┴──────────────────────────┴────────┴─────────┴───┘

The primary goal of this pull request is to allow `Z.chain` and `Z.chainRec` to handle very large arrays; the performance gains are a happy consequence of these functions receiving so much scrutiny. :)

index.js
safareli: Ah i see 👍

davidchambers:
What a wonderful team effort. Thanks, folks! :beers:

PR #112, opened 2019-02-17
derive ‘all’, ‘any’, and ‘none’ from ‘reduce’

davidchambers:
I've needed `all` and/or `any` several times during the past twelve months. These functions clearly belong in the standard library. `none` is useful for the same reason that `reject` is useful: it avoids double negatives.

I'd appreciate a review from a member of @sanctuary-js/owners. :)

/cc @abdul-takt, @adrienlo, @sphvn

### DIFF
--- a/index.js
+++ b/index.js
@@ -150,6 +150,13 @@
     };
   }
 
+  //  unary :: (a -> b) -> (a, Any...) -> b
+  function unary(f) {
+    return function(x) {
+      return f (x);
+    };
+  }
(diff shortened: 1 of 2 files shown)


adrienlo:
This is great! Thanks!

abdul-takt:
nice!

Bradcomp:
:bowtie:

sphvn:
Late to the party but 💯

PR #122, opened 2019-03-31
reformulate ‘takeWhile’ and ‘dropWhile’

davidchambers:
Commit message:

> Defining these functions in terms of `filter` is unsound: it relies on `fantasy-land/filter` enumeration order, which is unspecified.

[Gitter comment][1]:

> I noticed that making the implementation of `List#fantasy-land/filter` more efficient by reversing the list then filtering while reversing a second time actually broke `takeWhile` and `dropWhile`, as these functions rely on the order in which elements are filtered. Fantasy Land does not require that filtering happen in a particular order (nor should it). Furthermore, there are types (such as `Set a`) which can support `filter` but which cannot sensibly support `takeWhile` and `dropWhile`.

[1]: https://gitter.im/sanctuary-js/sanctuary?at=5c8feec08126720abc042faf

### DIFF
--- a/index.js
+++ b/index.js
@@ -1512,6 +1512,74 @@
     return filter (function(x) { return !(pred (x)); }, filterable);
   }
 
+  //# takeWhile :: (Applicative f, Foldable f, Monoid (f a)) => (a -> Boolean, f a) -> f a
+  //.
+  //. Discards the first element which does not satisfy the predicate, and all
+  //. subsequent elements.
+  //.
+  //. This function is derived from [`concat`](#concat), [`empty`](#empty),
+  //. [`of`](#of), and [`reduce`](#reduce).
+  //.
+  //. See also [`dropWhile`](#dropWhile).
+  //.
+  //. ```javascript
+  //. > takeWhile (s => /x/.test (s), ['xy', 'xz', 'yx', 'yz', 'zx', 'zy'])
+  //. ['xy', 'xz', 'yx']
+  //.
+  //. > takeWhile (s => /y/.test (s), ['xy', 'xz', 'yx', 'yz', 'zx', 'zy'])
+  //. ['xy']
+  //.
+  //. > takeWhile (s => /z/.test (s), ['xy', 'xz', 'yx', 'yz', 'zx', 'zy'])
+  //. []
+  //. ```
+  function takeWhile(pred, foldable) {
+    //  Fast path for arrays.
+    if (Array.isArray (foldable)) {
+      var idx = 0;
+      while (idx < foldable.length && pred (foldable[idx])) idx += 1;
+      return foldable.slice (0, idx);
+    }
+    var F = foldable.constructor;
+    return (reduce (function(acc, x) {
+      var take = acc.take && pred (x);
+      return {take: take, xs: take ? concat (acc.xs, of (F, x)) : acc.xs};
+    }, {take: true, xs: empty (F)}, foldable)).xs;
+  }
+
+  //# dropWhile :: (Applicative f, Foldable f, Monoid (f a)) => (a -> Boolean, f a) -> f a
+  //.
+  //. Retains the first element which does not satisfy the predicate, and all
+  //. subsequent elements.
+  //.
+  //. This function is derived from [`concat`](#concat), [`empty`](#empty),
+  //. [`of`](#of), and [`reduce`](#reduce).
+  //.
+  //. See also [`takeWhile`](#takeWhile).
+  //.
+  //. ```javascript
+  //. > dropWhile (s => /x/.test (s), ['xy', 'xz', 'yx', 'yz', 'zx', 'zy'])
+  //. ['yz', 'zx', 'zy']
+  //.
+  //. > dropWhile (s => /y/.test (s), ['xy', 'xz', 'yx', 'yz', 'zx', 'zy'])
+  //. ['xz', 'yx', 'yz', 'zx', 'zy']
+  //.
+  //. > dropWhile (s => /z/.test (s), ['xy', 'xz', 'yx', 'yz', 'zx', 'zy'])
+  //. ['xy', 'xz', 'yx', 'yz', 'zx', 'zy']
+  //. ```
+  function dropWhile(pred, foldable) {
+    //  Fast path for arrays.
+    if (Array.isArray (foldable)) {
+      var idx = 0;
+      while (idx < foldable.length && pred (foldable[idx])) idx += 1;
+      return foldable.slice (idx);
+    }
+    var F = foldable.constructor;
+    return (reduce (function(acc, x) {
+      var drop = acc.drop && pred (x);
+      return {drop: drop, xs: drop ? acc.xs : concat (acc.xs, of (F, x))};
+    }, {drop: true, xs: empty (F)}, foldable)).xs;
+  }
+
   //# map :: Functor f => (a -> b, f a) -> f b
   //.
   //. Function wrapper for [`fantasy-land/map`][].
@@ -2302,6 +2370,8 @@
     invert: invert,
     filter: filter,
     reject: reject,
+    takeWhile: takeWhile,
+    dropWhile: dropWhile,
     map: map,
     flip: flip,
     bimap: bimap,
--- a/test/index.js
+++ b/test/index.js
@@ -943,6 +943,42 @@ test ('reject', function() {
   eq (Z.reject (odd, Just (1)), Nothing);
 });
 
+test ('takeWhile', function() {
+  eq (Z.takeWhile.length, 2);
+  eq (Z.takeWhile.name, 'takeWhile');
+
+  eq (Z.takeWhile (odd, []), []);
+  eq (Z.takeWhile (odd, [1]), [1]);
+  eq (Z.takeWhile (odd, [1, 3]), [1, 3]);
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -2127,6 +2077,62 @@
...
+  //. ['xy']
+  //.
+  //. > takeWhile (s => /z/.test (s), ['xy', 'xz', 'yx', 'yz', 'zx', 'zy'])
+  //. []
+  //. ```
+  function takeWhile(pred, foldable) {
+    var F = foldable.constructor;
+    var take = true;
+    return reduce (function(xs, x) {
+      return ((take = take && pred (x))) ? concat (xs, of (F, x)) : xs;
```
masaeedu: I'm still a little bit uneasy about the `take = take && ...` side effect implicitly going on in here, although I can't formalize precisely why. I'll think about it a little bit.

davidchambers: It makes me uneasy too, but passing around `{ take :: Boolean, xs :: f a }` values is inelegant. Would you feel equally uneasy were `take` stored in the accumulator?

masaeedu: @davidchambers No, I think putting it into the zero for the fold and always pulling it off the `xs` for the fold would make me happy, because it's basically equivalent to a traversal wrt `State Boolean`.

masaeedu: I guess this is extremely pathological, but is the following implementation of `reduce` for arrays illegal wrt `Foldable`'s laws?

```
// :: (b -> a -> b) -> b -> [a] -> b
const reduce = f => z => xs => {
  const _ = xs.reduceRight((p, c) => f(p)(c), z)
  return xs.reduce((p, c) => f(p)(c), z)
}
```

davidchambers: ### Option :tada:: Variable reassignment

```javascript
function takeWhile(pred, foldable) {
  var F = foldable.constructor;
  var take = true;
  return reduce (function(xs, x) {
    return ((take = take && pred (x))) ? concat (xs, of (F, x)) : xs;
  }, empty (F), foldable);
}
```

### Option :heart:: Pure accumulator

```javascript
function takeWhile(pred, foldable) {
  var F = foldable.constructor;
  return (reduce (function(acc, x) {
    var take = acc.take && pred (x);
    return {take: take, xs: take ? concat (acc.xs, of (F, x)) : acc.xs};
  }, {take: true, xs: empty (F)}, foldable)).xs;
}
```

### Option :rocket:: Impure accumulator

```javascript
function takeWhile(pred, foldable) {
  var F = foldable.constructor;
  return (reduce (function(r, x) {
    r.xs = (r.take = r.take && pred (x)) ? concat (r.xs, of (F, x)) : r.xs;
    return r;
  }, {take: true, xs: empty (F)}, foldable)).xs;
}
```

Which option do you favour, Asad? Which option do others favour?

index.js
```diff
@@ -2127,6 +2077,62 @@
...
+  //. ['xz', 'yx', 'yz', 'zx', 'zy']
+  //.
+  //. > dropWhile (s => /z/.test (s), ['xy', 'xz', 'yx', 'yz', 'zx', 'zy'])
+  //. ['xy', 'xz', 'yx', 'yz', 'zx', 'zy']
+  //. ```
+  function dropWhile(pred, foldable) {
+    var F = foldable.constructor;
+    return (reduce (function(acc, x) {
+      var drop = acc.drop && pred (x);
+      return {drop: drop, xs: drop ? acc.xs : concat (acc.xs, of (F, x))};
```
masaeedu: This is cosmetic only, but are we allowed to use ES6 features like:

```
const drop = acc.drop && pred (x);
const xs = drop ? acc.xs : concat (acc.xs, of (F, x))

return { drop, xs }
```

?

davidchambers: All Sanctuary projects target ECMAScript 5.1, which I believe is the best version of the language that will ever exist.

I'm not trolling. :)

masaeedu: I have to get your reasoning on that at some point, but I guess that is a discussion for a different time. :smile:

davidchambers: I believe the proportion of well-considered language features whose virtues justify their complexity costs is decreasing as the language evolves. I believe this trend will continue. My extremely rough guess is that I currently approve of 4% of language features. ;)

JavaScript's core: `function f(x) { return x; }`, `(`, `)`, `[`, `]`, `{`, `}`, `'abc'`, `123`. Almost everything else can be implemented with these primitives.

masaeedu: I agree with the general sentiment, but I guess we differ on where we draw the line.

I don't view: `{ foo: 42, bar: 10 }` as a primitive of the language. Instead I view it as a sugar for `{ foo: 42, ...({ bar: 10, ...({}) }) }`, i.e. I see: `{} :: Object k v o` and `{ [k]: v, ...o } :: Object k v o` as the actual primitives of the language (similar for `[]` and `[x, ...xs]`).

As a result I think it's fine to use language features that pattern match and pun over `[]` and `{}` literals.

davidchambers: To be clear, the decision to target ECMAScript 5.1 is practical rather than philosophical. Were newer language features supported everywhere I would absolutely adopt a few. :)

index.js
gabejohnson: I prefer the variable reassignment for low-level utilities. There's not object creation overhead as there is for option :heart: and no property access overhead as there is for option :rocket:.

I don't have benchmarks to support my preference though.

I would never do this in application code unless it was on a hot path though 😄

gabejohnson:
@davidchambers have you considered special-casing `Array` and using `findIndex`&`slice` or a `for` loop?

index.js
davidchambers: Thanks for your input, Gabe. :)

Asad makes a good point, which is that it's not necessarily safe to provide an impure function to `reduce`. Consider this pathological case:

```javascript
Array.prototype['fantasy-land/reduce'] = function(f, x) {
  return this.reduce (f, x), this.reduce (f, x);
};
```

```javascript
var take = true;
Z.reduce ((xs, x) => (take = take && x <= 3) ? Z.concat (xs, [x]) : xs,
          [],
          [1, 2, 3, 4, 5]);
```

This will evaluate to `[]` rather than `[1, 2, 3]` because evaluating the first occurrence of `this.reduce (f, x)` will set `take` to `false`.

Is this `Array#fantasy-land/reduce` implementation sensible? No. It is lawful, though, so it should interoperate correctly with `Z` functions.

davidchambers:
Good idea, Gabe. I'll add a fast path for arrays. :)

index.js
```diff
@@ -2127,6 +2077,74 @@
...
+  //. > takeWhile (s => /z/.test (s), ['xy', 'xz', 'yx', 'yz', 'zx', 'zy'])
+  //. []
+  //. ```
+  function takeWhile(pred, foldable) {
+    //  Fast path for arrays.
+    if (Array.isArray (foldable)) {
+      var idx = 0;
+      while (idx < foldable.length && pred (foldable[idx])) idx += 1;
+      return foldable.slice (0, idx);
+    }
```
davidchambers: Do you like this implementation, @gabejohnson?

index.js
masaeedu: If we do want to do the efficient mutatey thing I'd be happier if it was an explicit traversal wrt `State` (not necessarily the pure `State`), because that wouldn't have these weird edge cases.

davidchambers: I don't know what you mean exactly, Asad. Are you saying you like :rocket: more than :tada:?

masaeedu: @davidchambers No, sorry, what I mean is that while :heart: is the option I like best of the ones you presented, I realize that it's inefficient.

If we want to use mutationto make this more efficient (while still preserving correctness), I would prefer if we captured this explicitly by using `traverse` wrt a state mutation effect.

This doesn't implicitly interleave side effects with the arbitrary typeclass implementations, the pathological scenario we were discussing earlier would still work correctly, and you still avoid any efficiency problems with the :heart: approach (whatever they may be, I haven't actually measured either).

davidchambers: Are you suggesting we use `traverse` rather than `reduce`? Could you suggest an implementation?

index.js
gabejohnson: That works, but you may be able to save a traversal by just copying over elements to a new array in the loop. Same time complexity though.

index.js
masaeedu: @davidchambers I tried implementing it but the best I could do was:

```
// :: Witherable f -> (a -> Boolean) -> f a -> f a
const takeWhile = W => test => {
  // :: a -> State Boolean (Maybe a)
  const action = a => stop =>
    stop || !test(a) ? [Nothing, false] : [Just(a), true];

  return Fn.pipe([
    W.wither(State)(action),
    State.eval
  ]);
};
```

where `type Witherable t = (Traversable t) & { wither: Applicative f -> (a -> f (Maybe b)) -> t a -> f (t b) }`. I feel like there might be a way to do it using only `Traversable` but I couldn't quite figure it out.

While writing this, I looked back at the implementation we have now and started wondering what would happen if I tried to `takeWhile` a `ZipList` or an `Observable` (where the `of` produces an infinite "repetition" of the value). We should probably try out the implementation on a few different sorts of things, or if we only plan to support arrays we should just specialize it to arrays.

davidchambers:
> We should probably try out the implementation on a few different sorts of things, or if we only plan to support arrays we should just specialize it to arrays.

We specialized Sanctuary's implementations in sanctuary-js/sanctuary#619. The generalizations proposed in this pull request are problematic; we should stick with the array-specific functions.

PR #123, opened 2019-04-04
remove ‘takeWhile’ and ‘dropWhile’

davidchambers:
These functions are unsound, as explained in #122. Given that it may take us a long time to reformulate these functions, I propose the following actions:

1.  Remove `Z.takeWhile` and `Z.dropWhile`.
2.  Release `sanctuary-type-classes@11.0.0`.
3.  Specialize `S.takeWhile` and `S.dropWhile` to `(a -> Boolean) -> Array a -> Array a`.
4.  Release `sanctuary@2.0.0`.
5.  Take as much time as necessary to resolve #122 satisfactorily.

There will be one breaking change to each project; this two-phase approach is no more disruptive than merging #122 immediately would have been.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1512,56 +1512,6 @@
     return filter (function(x) { return !(pred (x)); }, filterable);
   }
 
-  //# takeWhile :: Filterable f => (a -> Boolean, f a) -> f a
-  //.
-  //. Discards the first element which does not satisfy the predicate, and all
-  //. subsequent elements.
-  //.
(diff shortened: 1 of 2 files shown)


masaeedu:
Makes sense to me. Does anyone know of important non-array usages of takeWhile that would be broken by this?

gabejohnson:
@davidchambers I think this is the best option ATM.

PR #128, opened 2019-11-03
sanctuary-type-identifiers@3.0.0

davidchambers:
<https://github.com/sanctuary-js/sanctuary-type-identifiers/releases/tag/v3.0.0>

We wish to update and release sanctuary-type-classes before updating the various ADT packages, because they depend on this package. As a result, this project's test suite will temporarily rely on the fact that `Object#fantasy-land/equals` happens to correctly determine equality of `Identity a`, `Maybe a`, and `Pair a b` values.

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -319,7 +319,7 @@ test ('Setoid', function() {
   eq (Z.Setoid.test (''), true);
   eq (Z.Setoid.test ([]), true);
   eq (Z.Setoid.test ({}), true);
-  eq (Z.Setoid.test ({'@@type': 'my-package/Quux@1'}), true);
+  eq (Z.Setoid.test ({'@@type': 'my-package/Quux@1'}), false);
 });
 
 test ('Ord', function() {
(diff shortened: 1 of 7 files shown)


test/index.js
```diff
@@ -668,8 +668,9 @@ test ('equals', function() {
   eq (Z.equals (Math.sin, Math.cos), false);
   eq (Z.equals (Identity (Identity (Identity (0))), Identity (Identity (Identity (0)))), true);
   eq (Z.equals (Identity (Identity (Identity (0))), Identity (Identity (Identity (1)))), false);
-  eq (Z.equals ({'@@type': 'my-package/Quux@1'}, {'@@type': 'my-package/Quux@1'}), true);
+  eq (Z.equals ({'@@type': 'my-package/Quux@1'}, {'@@type': 'my-package/Quux@1'}), false);
   eq (Z.equals (Array.prototype, Array.prototype), true);
+  delete Maybe['@@type'];
```
davidchambers: This `delete` statement is required to prevent `Maybe` from being treated as a member of a type that lacks a `fantasy-land/equals` method. I will remove it when I upgrade the sanctuary-maybe dependency.

davidchambers: #135

PR #133, opened 2020-01-16
derive ‘intercalate’ from ‘concat’, ‘empty’, and ‘reduce’

davidchambers:
@syaiful6 first proposed this function in March 2017 (sanctuary-js/sanctuary#360). @masaeedu opened a pull request to add the function in June 2018 (sanctuary-js/sanctuary#569). Let's finally define `intercalate` in January 2020. :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -2036,6 +2036,47 @@
     return any (function(y) { return equals (x, y); }, foldable);
   }
 
+  //# intercalate :: (Monoid m, Foldable f) => (m, f m) -> m
+  //.
+  //. Concatenates the elements of the given structure, separating each pair
+  //. of adjacent elements with the given separator.
+  //.
+  //. This function is derived from [`concat`](#concat), [`empty`](#empty),
+  //. and [`reduce`](#reduce).
+  //.
+  //. ```javascript
+  //. > intercalate (', ', [])
+  //. ''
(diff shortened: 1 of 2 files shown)


davidchambers:
> Will it replace <https://sanctuary.js.org/#joinWith> eventually?

Good question. I like having `S.joinWith (s)` as the inverse of `S.splitOn (s)`, and I'm not opposed to including a specialization of another Sanctuary function if its type signature is significantly simpler. In fact, I believe we should add your [combinators][1] to Sanctuary: although duplicative, they would provide significant pedagogical value and (soon) useful search results.


[1]: https://gist.github.com/Avaq/1f0636ec5c8d6aed2e45

Avaq:
> they would provide significant pedagogical value and (soon) useful search results

That might work. But do we really want to increase the API surface to improve search results? Maybe for the website we could implement some sort of "search-only aliases".

- We have the combinators that map to several existing functions
- We have the Ramda dictionary which maps to Sanctuary functions
- And now we have specialized functions which have been generalized and renamed

All of these could make use of a search-only mapping, where if a user types `unnest` or `flatten` they might find `join`, and if they type `String -> Array String -> String` or `joinWith` they might find `intercalate`.

PR #142, opened 2020-08-19
remove Function.name polyfill

davidchambers:
[`Function.name`][1] has been widely supported for years at this point, so the polyfill no longer seems necessary.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name

### DIFF
--- a/index.js
+++ b/index.js
@@ -254,15 +254,6 @@
     return _funcPath (false, path, implementations);
   }
 
-  //  functionName :: Function -> String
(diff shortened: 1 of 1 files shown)


Avaq:
> Perhaps this pull request should wait a few years. What do others think?

Since the name is central to the design of this lib, I guess it'd be good to keep the polyfill around until we're certain it's no longer needed.

davidchambers:
Let's revisit this in a few years. :)

PR #146, opened 2021-04-19
refactor implementation lookup

davidchambers:
This pull request removes the awful `funcPath` function, and splits `getBoundMethod` into `getStaticMethod` and `getPrototypeMethod` (which can be top-level functions as they do not reference `requirements`).

Aldwin and I have made changes to these several times over the years. :joy:

### DIFF
--- a/index.js
+++ b/index.js
@@ -233,14 +233,6 @@
   //  Value :: Location
   var Value = 'Value';
 
-  //  funcPath :: (Array String, a) -> Nullable Function
-  function funcPath(path, _x) {
-    var x = _x;
-    // eslint-disable-next-line no-plusplus
-    for (var idx = 0; x != null && idx < path.length; x = x[path[idx++]]);
-    return typeof x === 'function' ? x : null;
-  }
-
   //  functionName :: Function -> String
   var functionName = has ('name', function f() {}) ?
     function functionName(f) { return f.name; } :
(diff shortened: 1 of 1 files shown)


Avaq (approved):

index.js
```diff
@@ -250,47 +242,77 @@
...
-      methods[_name] = getBoundMethod (_name);
-      return methods;
-    }, {});
+    typeClass.methods = {};
+    staticMethodNames.forEach (function(_name) {
+      typeClass.methods[_name] = getStaticMethod (_name);
+    });
+    prototypeMethodNames.forEach (function(_name) {
+      typeClass.methods[_name] = function(x) {
+        return (getPrototypeMethod (_name) (x)).bind (x);
```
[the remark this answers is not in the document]
Avaq: Yeah, see also 4c9ddd72f5d81c91f0bf7f271ae875486a085c88

index.js
```diff
@@ -250,47 +242,77 @@
...
       dependencies,
       function(x) {
-        return keys.every (function(_name) {
-          var arg = requirements[_name] === Constructor ? x.constructor : x;
-          return getBoundMethod (_name) (arg) != null;
-        });
+        return (
+          staticMethodNames.every (function(_name) {
+            return x != null &&
+                   getStaticMethod (_name) (x.constructor) != null;
```
davidchambers, line 298: The test suite passes without the `x != null` guard. All of this library's type classes that require static methods happen to depend on type classes that require prototype methods that `Null` and `Undefined` cannot provide.

These steps demonstrate that `x.constructor` should be guarded:

1.  Remove the `x != null` guard.
2.  Remove the `Semigroupoid` dependency from `Category`.
3.  Evaluate `Z.Category.test (null)`.

Since users can define their own type classes via `Z.TypeClass`, it is necessary to include the guard.

Avaq: > users can define their own type classes [which necessitate inclusion of] the guard

That sounds like it might warrant a test.

PR #147, opened 2021-04-21
Remove special treatment of non-compliant values in lte and equals

Avaq:
This PR competes with and so closes #154

Instead of arbitrarily returning false, these functions now throw when
given invalid (without the proper type class) input.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1111,10 +1111,11 @@
     }
   };
 
-  //# equals :: (a, b) -> Boolean
+  //# equals :: Setoid a => (a, a) -> Boolean
   //.
   //. Returns `true` if its arguments are of the same type and equal according
-  //. to the type's [`fantasy-land/equals`][] method; `false` otherwise.
+  //. to the type's [`fantasy-land/equals`][] method. Returns `false` for
+  //. inputs of differing types.
   //.
   //. `fantasy-land/equals` implementations are provided for the following
   //. built-in types: Null, Undefined, Boolean, Number, Date, RegExp, String,
@@ -1154,20 +1155,18 @@
 
       $pairs.push ([x, y]);
       try {
-        return Z.Setoid.test (x) &&
-               Z.Setoid.test (y) &&
-               Z.Setoid.methods.equals (x) (y);
+        return Z.Setoid.methods.equals (x) (y);
       } finally {
         $pairs.pop ();
       }
     };
   }
 
-  //# lt :: (a, b) -> Boolean
+  //# lt :: Ord a => (a, a) -> Boolean
   //.
   //. Returns `true` if its arguments are of the same type and the first is
   //. less than the second according to the type's [`fantasy-land/lte`][]
-  //. method; `false` otherwise.
+  //. method. Returns `false` for inputs of differing types.
   //.
   //. This function is derived from [`lte`](#lte).
   //.
@@ -1185,11 +1184,12 @@
   //. ```
   Z.lt = (x, y) => sameType (x, y) && !(Z.lte (y, x));
 
-  //# lte :: (a, b) -> Boolean
+  //# lte :: Ord a => (a, a) -> Boolean
   //.
   //. Returns `true` if its arguments are of the same type and the first
   //. is less than or equal to the second according to the type's
-  //. [`fantasy-land/lte`][] method; `false` otherwise.
+  //. [`fantasy-land/lte`][] method. Returns `false` for inputs of
+  //. differing types.
   //.
   //. `fantasy-land/lte` implementations are provided for the following
   //. built-in types: Null, Undefined, Boolean, Number, Date, String, Array,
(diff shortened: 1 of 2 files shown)


davidchambers:
We should update the type signatures of `equals`, `lt`, `lte`, `gt`, and `gte`. :)

Avaq:
In which way?

Avaq:
Oh! I see. Yes.

Avaq:
I can't figure out the signature. Hindley Milner is not versatile enough to cleanly separate garbage inputs from desired inputs. The best I can come up with is this overloaded function.

```hs
equals :: Setoid a => (a, a) -> Boolean
equals :: (a, AnythingBut a) -> Boolean
```

Avaq:
We could say that the function wants `(Setoid a, Setoid b) => (a, b) -> Boolean`, but that precludes non-setoid values of differing types.

davidchambers:
I suggest using `equals :: Setoid a => (a, a) -> Boolean`. A clear, imprecise type signature has more pedagogical value than an unclear, precise one. We can document in other ways how the function behaves when given values of different types.

Avaq:
@davidchambers Updated.

index.js
```diff
@@ -1176,10 +1176,11 @@
     }
   }
 
-  //# equals :: (a, b) -> Boolean
+  //# equals :: Setoid a => (a, a) -> Boolean
   //.
   //. Returns `true` if its arguments are of the same type and equal according
-  //. to the type's [`fantasy-land/equals`][] method; `false` otherwise.
+  //. to the type's [`fantasy-land/equals`][] method. Returns `false` for
+  //. inputs of differing types.
```
davidchambers, line 1183: I don't think this change is necessary, as we already include the clause _if its arguments are of the same type_.

test/index.js
```diff
@@ -675,11 +675,21 @@ test ('equals', function() {
...
   eq (Z.equals (Array.prototype, Array.prototype), true);
   eq (Z.equals (Nothing.constructor, Maybe), true);
   eq (Z.equals ((Just (0)).constructor, Maybe), true);
-  eq (Z.equals (Lazy$of (0), Lazy$of (0)), false);
+
+  assert.throws (() => {
+    Z.equals (Useless, Useless);
+  }, (
+    new TypeError ("Cannot read property 'bind' of null")
+  ));
```
davidchambers, line 686: I would prefer this formatting:

```javascript
  assert.throws (
    () => { Z.equals (Useless, Useless); },
    new TypeError ("Cannot read property 'bind' of null")
  );
```

index.js
Avaq: But `false otherwise` is too broad: When a user provides two values of the same type that don't have Setoid, the function throws.

davidchambers: We do not typically document the behaviour of a function when applied to an argument outside its domain. The description of [`S.trim`][1], for example, does not consider the possibility of application to a non-string value.


[1]: https://sanctuary.js.org/#trim

Avaq: So you'd prefer to remove the *`false` otherwise* part altogether, and treat the special treatment of inputs of differing types as an undocumented "feature", or simply garbage?

test/index.js
```diff
@@ -675,11 +675,19 @@ test ('equals', function() {
...
   eq (Z.equals (Identity (Identity (Identity (0))), Identity (Identity (Identity (1)))), false);
-  eq (Z.equals (Useless, Useless), false);
   eq (Z.equals (Array.prototype, Array.prototype), true);
   eq (Z.equals (Nothing.constructor, Maybe), true);
   eq (Z.equals ((Just (0)).constructor, Maybe), true);
-  eq (Z.equals (Lazy$of (0), Lazy$of (0)), false);
+
+  assert.throws (
+    () => { Z.equals (Useless, Useless); },
+    new TypeError ("Cannot read property 'bind' of null")
```
davidchambers, line 684: Did you consider detecting this case and providing a more informative error message? Mentioning `fantasy-land/equals` here would be helpful.

index.js
Avaq: > when applied to an argument outside its domain

Supplying two values of differing types are within the domain of the function. So are supplying two Setoids of the same type. Anything else isn't.

Avaq: So I've documented purely the behaviour of the function for values *within* its domain. I have removed documentation that overlaps with values outside of the domain.

test/index.js
Avaq: Yes. I intended to make that change in a follow-up PR. I prefer to consider the perf PR first though, because the code we'd need to touch gets reworked there.

Avaq: The reason this test case is listed is because it used to be a valid input. For most other functions, the garbage paths are not tested. Perhaps I should remove this test case altogether.

Avaq: Actually it's silly to test "unspecified behaviour", because that makes it specified. :P

Avaq: I've opened #152 to track your suggestion. I'll also remove this test case from the PR.

index.js
davidchambers: I do not see how `Setoid a => (a, a) -> Boolean` permits _inputs of differing types_. :thinking:

Avaq: But the *real* signature is:

```hs
equals :: Setoid a => (a, a) -> Boolean
equals :: (a, AnythingBut a) -> Boolean
```

Remember..? But you said that is not pedagogical.

Avaq: The function has a special code path to facilitate the latter overload.

Avaq: That makes these inputs part of the domain in my book.

Avaq: You yourself said:

> We can document in other ways how the function behaves when given values of different types. -- https://github.com/sanctuary-js/sanctuary-type-classes/pull/147#issuecomment-824666181

That's what I'm trying to do here.

PR #154, opened 2021-04-24
Update the equals function to fall back to structural equality

Avaq:
This PR competes with and so closes #147 

When comparing two values that are:

1. of the same type; and
2. do not provide a 'fantasy-land/equals'-method; but
3. also don't have a known type with a Z-provided implementation,

then previously Z.equals would return false, but this false is just as
reliable as a true would be.

After this PR, instead of returning false, the Z.equals now returns
the structural equality of the two inputs.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1113,12 +1113,20 @@
 
   //# equals :: (a, b) -> Boolean
   //.
-  //. Returns `true` if its arguments are of the same type and equal according
-  //. to the type's [`fantasy-land/equals`][] method; `false` otherwise.
+  //. Returns `true` if its arguments are equal; `false` otherwise.
   //.
-  //. `fantasy-land/equals` implementations are provided for the following
-  //. built-in types: Null, Undefined, Boolean, Number, Date, RegExp, String,
-  //. Array, Arguments, Error, Object, and Function.
+  //. Specifically:
+  //.
+  //.   - Arguments with different [type identities][] are unequal.
+  //.
+  //.   - If the first argument has a [`fantasy-land/equals`][] method,
+  //.     that method is invoked to determine whether the arguments are
+  //.     equal (`fantasy-land/equals` implementations are provided for the
+  //.     following built-in types: Null, Undefined, Boolean, Number, Date,
+  //.     RegExp, String, Array, Arguments, Error, Object, and Function).
+  //.
+  //.   - Otherwise, the arguments are equal if their
+  //.     [entries][`Object.entries`] are equal (according to this algorithm).
   //.
   //. The algorithm supports circular data structures. Two arrays are equal
   //. if they have the same index paths and for each path have equal values.
@@ -1154,9 +1162,9 @@
 
       $pairs.push ([x, y]);
       try {
-        return Z.Setoid.test (x) &&
-               Z.Setoid.test (y) &&
-               Z.Setoid.methods.equals (x) (y);
+        return Z.Setoid.test (x) ?
+          Z.Setoid.methods.equals (x) (y) :
+          Object$prototype$equals.call (x, y);
       } finally {
         $pairs.pop ();
       }
@@ -2253,6 +2261,7 @@
 //. [Semigroupoid]:             v:fantasyland/fantasy-land#semigroupoid
 //. [Setoid]:                   v:fantasyland/fantasy-land#setoid
 //. [Traversable]:              v:fantasyland/fantasy-land#traversable
+//. [`Object.entries`]:         https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries
 //. [`fantasy-land/alt`]:       v:fantasyland/fantasy-land#alt-method
 //. [`fantasy-land/ap`]:        v:fantasyland/fantasy-land#ap-method
 //. [`fantasy-land/bimap`]:     v:fantasyland/fantasy-land#bimap-method
@@ -2276,4 +2285,5 @@
 //. [`fantasy-land/traverse`]:  v:fantasyland/fantasy-land#traverse-method
 //. [`fantasy-land/zero`]:      v:fantasyland/fantasy-land#zero-method
 //. [stable sort]:              https://en.wikipedia.org/wiki/Sorting_algorithm#Stability
+//. [type identities]:          v:sanctuary-js/sanctuary-type-identifiers
 //. [type-classes]:             https://github.com/sanctuary-js/sanctuary-def#type-classes
--- a/test/index.js
+++ b/test/index.js
@@ -567,6 +567,33 @@ test ('Contravariant', () => {
 });
 
 test ('equals', () => {
+  const {nestedSetoidArb} = jsc.letrec (tie => ({
+    builtinSetoidArb: jsc.record ({
+      value: tie ('nestedSetoidArb'),
+    }),
+    noSetoidArb: jsc.record ({
+      'value': tie ('nestedSetoidArb'),
+      '@@type': jsc.constant ('sanctuary-type-classes/NoSetoid@1'),
+      '@@show': jsc.constant (function() { return `NoSetoid (${show (this.value)})`; }),
+    }),
+    customSetoidArb: jsc.record ({
+      'value': tie ('nestedSetoidArb'),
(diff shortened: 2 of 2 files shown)


Avaq:
I'm looking into what it means, conceptually, to make this same change to `Z.lte`.

index.js
```diff
@@ -1219,9 +1223,9 @@
 
       $pairs.push ([x, y]);
       try {
-        return Setoid.test (x) &&
-               Setoid.test (y) &&
```
Avaq, line 1223: This was actually a low-hanging performance improvement that I think we missed. There's no need to Setoid.test the second value, because it will never be asked to provide a `fantasy-land/equals` method, and we already predetermined that the two values are of the same type.

Avaq: latest-screenshot

Avaq: That's compared to `master`.

test/index.js
```diff
@@ -675,7 +675,7 @@ test ('equals', function() {
   eq (Z.equals (Math.sin, Math.cos), false);
   eq (Z.equals (Identity (Identity (Identity (0))), Identity (Identity (Identity (0)))), true);
   eq (Z.equals (Identity (Identity (Identity (0))), Identity (Identity (Identity (1)))), false);
-  eq (Z.equals (Useless, Useless), false);
+  eq (Z.equals (Useless, Useless), true);
```
Avaq, line 678: Take note that `Z.equals (Lazy$of (0), Lazy$of (0))` is still `false`. But now:

```js
const x = Lazy$of (0);
Z.equals (x, x)
```

should be `true`

Avaq:
I prefer this approach over the direction #147 is going. How do you feel about merging this @davidchambers ?

davidchambers:
> How do you feel about merging this @davidchambers ?

I am happy with the new behaviour. I would like to spend some time reworking the function's description.

index.js
```diff
@@ -1154,9 +1158,9 @@
 
       $pairs.push ([x, y]);
       try {
-        return Z.Setoid.test (x) &&
-               Z.Setoid.test (y) &&
-               Z.Setoid.methods.equals (x) (y);
+        return Z.Setoid.test (x) ?
+               Z.Setoid.methods.equals (x) (y) :
+               Object$prototype$equals.call (x, y);
```
davidchambers, line 1163: This line wrapping would be more consistent with the existing code:

```suggestion
        return Z.Setoid.test (x) ?
          Z.Setoid.methods.equals (x) (y) :
          Object$prototype$equals.call (x, y);
```

index.js
```diff
@@ -1113,12 +1113,16 @@
...
-  //. `fantasy-land/equals` implementations are provided for the following
-  //. built-in types: Null, Undefined, Boolean, Number, Date, RegExp, String,
-  //. Array, Arguments, Error, Object, and Function.
+  //. - The types of the two values. Values of differing types unequal.
+  //. - The [`fantasy-land/equals`][] method of the first value when present.
+  //. - For Null, Undefined, Boolean, Number, Date, RegExp, String, Array,
+  //.   Arguments, Error, Object, and Function, Sanctuary Type Classes provides
+  //.   `fantasy-land/equals` implementations.
+  //. - Any other types are structurally compared by recursively comparing the
+  //.   values at each of the keys.
```
davidchambers, line 1125: I propose the following description:

```javascript
  //. Returns `true` if its arguments are equal; `false` otherwise.
  //.
  //. Specifically:
  //.
  //.   - Arguments of different types are unequal.
  //.
  //.   - If the first argument has a [`fantasy-land/equals`][] method,
  //.     that method is invoked to determine whether the arguments are
  //.     equal (`fantasy-land/equals` implementations are provided for the
  //.     following built-in types: Null, Undefined, Boolean, Number, Date,
  //.     RegExp, String, Array, Arguments, Error, Object, and Function).
  //.
  //.   - Otherwise, the arguments are equal if their
  //.     [entries][`Object.entries`] are equal (according to this algorithm).
```

```javascript
//. [`Object.entries`]:         https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries
```

index.js
```diff
@@ -2275,5 +2283,6 @@
 //. [`fantasy-land/reduce`]:    v:fantasyland/fantasy-land#reduce-method
 //. [`fantasy-land/traverse`]:  v:fantasyland/fantasy-land#traverse-method
 //. [`fantasy-land/zero`]:      v:fantasyland/fantasy-land#zero-method
+//. [`Object.entries`]:         https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries
```
davidchambers, line 2286: I use case-sensitive ordering; running `:sort` in Vim does most of the work for me.

Please insert this entry between the `Traversable` and `` `fantasy-land/alt` `` entries. :)

Avaq: Ah, I tried sorting with vscode, and it completely changed the order of everything, so I assumed items were grouped.

test/index.js
```diff
@@ -567,6 +567,31 @@ test ('Contravariant', () => {
...
+    }),
+    nestedSetoidArb: jsc.oneof ([
+      tie ('builtinSetoidArb'),
+      tie ('noSetoidArb'),
+      tie ('customSetoidArb'),
+      jsc.nat,
+    ]),
+  }));
+
+  nestedSetoidArb.show = show;
```
davidchambers, line 593: What does this line do?

test/index.js
```diff
@@ -686,6 +711,8 @@ test ('equals', () => {
   eq (Z.equals ($0, $0), true);
   eq (Z.equals ($0, $1), false);
   eq (Z.equals ($1, $0), false);
+
+  jsc.assert (jsc.forall (nestedSetoidArb, x => Z.equals (x, x)));
```
davidchambers, line 715: Clever approach! :clap:

test/index.js
```diff
@@ -567,6 +567,31 @@ test ('Contravariant', () => {
...
+    noSetoidArb: jsc.record ({
+      'a': tie ('nestedSetoidArb'),
+      '@@type': jsc.constant ('sanctuary-type-classes/NoSetoid@1'),
+      '@@show': jsc.constant (function() { return `NoSetoid (${show (this.a)})`; }),
+    }),
+    customSetoidArb: jsc.record ({
+      'a': tie ('nestedSetoidArb'),
+      '@@type': jsc.constant ('sanctuary-type-classes/CustomSetoid@1'),
+      '@@show': jsc.constant (function() { return `CustomSetoid (${show (this.a)})`; }),
+      'fantasy-land/equals': jsc.constant (function(that) { return Z.equals (this.a, that.a); }),
```
davidchambers, line 583: I would like to rename the parameter `other`. Also, let's rename the `a` field `value`.

test/index.js
Avaq: It overrides the built-in `show` method provided by jsverify, which is just JSON.stringify, so that if an error occurs, the output shows the two inequal values nicely.

index.js
```diff
@@ -1113,12 +1113,20 @@
...
-  //. Returns `true` if its arguments are of the same type and equal according
-  //. to the type's [`fantasy-land/equals`][] method; `false` otherwise.
+  //. Returns `true` if its arguments are equal; `false` otherwise.
   //.
-  //. `fantasy-land/equals` implementations are provided for the following
-  //. built-in types: Null, Undefined, Boolean, Number, Date, RegExp, String,
-  //. Array, Arguments, Error, Object, and Function.
+  //. Specifically:
+  //.
+  //.   - Arguments of different types are unequal.
```
Avaq, line 1120: Should we mention that we're using sanctuary-type-identifiers to differentiate types?

davidchambers: Yes, that's a good idea.

Avaq: How about *Arguments with different [type identities](https://github.com/sanctuary-js/sanctuary-type-identifiers) are unequal*?

davidchambers: Excellent! We can use `v:sanctuary-js/sanctuary-type-identifiers` to link to the right version. :)

davidchambers (approved):
Thank you for persevering with this!

PR #155, opened 2021-04-24
Update the lte function to fall back to structural ordering

Avaq:
This is the counterpart to #154. Just opened it for consideration.


test/index.js
```diff
@@ -802,7 +802,7 @@ test ('lte', function() {
   eq (Z.lte (Identity (Identity (Identity (0))), Identity (Identity (Identity (0)))), true);
   eq (Z.lte (Identity (Identity (Identity (0))), Identity (Identity (Identity (1)))), true);
   eq (Z.lte (Identity (Identity (Identity (1))), Identity (Identity (Identity (0)))), false);
-  eq (Z.lte (Lazy$of (0), Lazy$of (0)), false);
+  eq (Z.lte (Lazy$of (0), Lazy$of (0)), true);
```
Avaq, line 805: I believe this started returning `true` because there is no `Function$prototype$ord` implementation, so two functions will be structurally compared: `Z.lte (() => {}, () => {})` now evaluates to `true` because the two values are structurally "less than or equal".

Oddly, in #154, `Z.equals (() => {}, () => {})` evaluates to `false`, because we have an implementation of Setoid for functions that uses reference equality.

Avaq: I would expect that the following always holds: `Z.equals (a, b) && Z.lte (a, b) == Z.equals (a, b)`, but this rule is now broken.

Avaq:
Closing in favour of #163

PR #151, opened 2021-04-22
Performance improvements

Avaq:
Together with #147, this brings an overall performance improvement across the library. Compare performance across commits by checking the commit messages.

PR #156, opened 2021-04-24
use ES6 syntax

davidchambers:
sanctuary-js/sanctuary#687

### DIFF
--- a/index.js
+++ b/index.js
@@ -68,7 +68,7 @@
 //.
 //. ## API
 
-(function(f) {
+(f => {
 
   'use strict';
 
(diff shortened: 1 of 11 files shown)


index.js
```diff
@@ -81,90 +81,50 @@
...
 
-} (function(type) {
+}) (type => {
 
   'use strict';
 
   /* istanbul ignore if */
   if (typeof __doctest !== 'undefined') {
-    /* eslint-disable no-unused-vars */
+    /* eslint-disable no-unused-vars, no-var */
```
Avaq, line 90: Interesting. I never thought to use `var` as a feature. :)

index.js
```diff
@@ -212,71 +172,64 @@
...
 
   //  Value :: Location
-  var Value = 'Value';
+  const Value = 'Value';
 
   //  functionName :: Function -> String
-  var functionName = has ('name', Function) ?
-    function functionName(f) { return f.name; } :
+  const functionName = (
+    has ('name', Function) ?
```
Avaq, line 192: Do we still need to consider the fallback for function names? Wasn't that an es5-support thing?

davidchambers: Good point! You are correct: [ECMAScript 2015 Language Specification § 19.2.4.2][1].


[1]: https://262.ecma-international.org/6.0/#sec-function-instances-name

davidchambers: #142 :thinking:

Avaq: Yeah, and depending on the outcome of this https://github.com/sanctuary-js/sanctuary-type-classes/pull/151#discussion_r618284535 discussion, we may end up removing the `functionName` function altogether.

Avaq (approved):

PR #160, opened 2021-05-01
Fix incorrect usage of arb.smap in tests

Avaq:
As I understand, the .smap method is not a map function. Its intended use
is to provide a way for JSVerify to verify that, given a set, a reduction
of that set through a given function will produce a succesful assertion if
the greater set fails the same assertion.

It takes two functions: f, and g, where g should be the right inverse of
f. The current code only passes f, and leaves g to undefined. This means
that if JSVerify would ever find a failing property using one of these
incorrectly smapped arbitraries, and deem it necessary to attempt
shrinking it, it would throw an error like: 'g is not a function'.

See also https://github.com/jsverify/jsverify/issues/104

This commit makes two changes:

1. A 'smapArb' dispatcher was added which forces both arguments to be
   provided. All occurrences of the arb.smap call where a right inverse
   function could be defined have been replaced to use this new
   dispatcher.
2. A 'mapArb' dispatcher was added which can be used safely in the way
   that arb.smap was previously used. It relies on jsc.bless to attach a
   default shrink function. The default shrink function just produces an
   empty set.

<sub>/cc @dicearr (him and I have been bitten by this same mistake many times in the past)</sub>

### DIFF
--- a/test/index.js
+++ b/test/index.js
@@ -7,6 +7,7 @@ const jsc = require ('jsverify');
 const Identity = require ('sanctuary-identity');
 const Maybe = require ('sanctuary-maybe');
 const Pair = require ('sanctuary-pair');
+const show = require ('sanctuary-show');
 const type = require ('sanctuary-type-identifiers');
 const Useless = require ('sanctuary-useless');
 
@@ -23,6 +24,12 @@ const {withUnstableArraySort} = require ('./quicksort');
 const {Nil, Cons} = List;
 const {Nothing, Just} = Maybe;
 
+//  mapArb :: (a -> b) -> Arbitrary a -> Arbitrary b
+const mapArb = f => arb => jsc.bless ({generator: arb.generator.map (f)});
+
+//  smapArb :: (a -> b) -> (b -> a) -> Arbitrary a -> Arbitrary b
+const smapArb = f => g => arb => arb.smap (f, g, show);
+
 //  Lazy$of :: a -> Lazy a
 function Lazy$of(x) {
   eq (arguments.length, Lazy$of.length);
@@ -30,17 +37,10 @@ function Lazy$of(x) {
 }
 
 //  ListArb :: Arbitrary a -> Arbitrary (List a)
-const ListArb = arb => (
-  jsc.array (arb)
-  .smap (xs => {
-    let list = Nil;
-    for (let idx = 0; idx < xs.length; idx += 1) list = Cons (xs[idx], list);
-    return list;
-  })
-);
+const ListArb = arb => smapArb (arrayToList) (listToArray) (jsc.array (arb));
 
 //  MaybeArb :: Arbitrary a -> Arbitrary (Maybe a)
-const MaybeArb = arb => jsc.oneof (jsc.constant (Nothing), arb.smap (Just));
+const MaybeArb = arb => jsc.oneof (jsc.constant (Nothing), mapArb (Just) (arb));
 
 //  Point :: () -> Point
 function Point() {
@@ -76,6 +76,13 @@ function args() {
   // eslint-disable-next-line prefer-rest-params
   return arguments;
 }
+
+//  arrayToList :: Array a -> List a
+const arrayToList = xs => {
+  let list = Nil;
+  for (let idx = 0; idx < xs.length; idx += 1) list = Cons (xs[idx], list);
+  return list;
+};
 
 //  compose :: (b -> c) -> (a -> b) -> a -> c
 function compose(f) {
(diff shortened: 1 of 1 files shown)


davidchambers:
> This means that if JSVerify would ever find a failing property using one of these incorrectly smapped arbitraries, and deem it necessary to attempt shrinking it, it would throw an error like: 'g is not a function'.

I have encountered this many times! Thank you for fixing the underlying issue. :clap:

test/index.js
```diff
@@ -23,24 +24,23 @@ const {withUnstableArraySort} = require ('./quicksort');
 const {Nil, Cons} = List;
 const {Nothing, Just} = Maybe;
 
+//    mapArb :: (a -> b) -> Arbitrary a -> Arbitrary b
```
davidchambers, line 27:
```suggestion
//  mapArb :: (a -> b) -> Arbitrary a -> Arbitrary b
```

test/index.js
```diff
@@ -23,24 +24,23 @@ const {withUnstableArraySort} = require ('./quicksort');
...
+const mapArb = f => arb => jsc.bless ({generator: arb.generator.map (f)});
+
+//    smapArb :: (a -> b) -> (b -> a) -> Arbitrary a -> Arbitrary b
```
davidchambers, line 30:
```suggestion
//  smapArb :: (a -> b) -> (b -> a) -> Arbitrary a -> Arbitrary b
```

test/index.js
```diff
@@ -77,6 +77,13 @@ function args() {
   return arguments;
 }
 
+//    arrayToList :: Array a -> List a
```
davidchambers, line 80:
```suggestion
//  arrayToList :: Array a -> List a
```

test/index.js
```diff
@@ -1652,15 +1659,19 @@ suite ('laws', () => {
   });
 
   suite ('RegExp', () => {
-    const arb = jsc.string.smap (s => {
-      const validFlags = ['', 'g', 'i', 'm', 'gi', 'gm', 'im', 'gim'];
-      const flags = validFlags[Math.floor (Math.random () * validFlags.length)];
```
Avaq, line 1657: While I was at it, I also removed this problematic code. One of the benefits of prop tests is that you can reproduce a failing assertion by rerunning the tests with the same seed. Of course this depends on your arbitraries being defined in terms of pure functions.

test/index.js
```diff
@@ -1652,15 +1659,19 @@ suite ('laws', () => {
...
+    const flagsArb = jsc.oneof (['', 'g', 'i', 'm', 'gi', 'gm', 'im', 'gim'].map (jsc.constant));
+    const patternArb = jsc.suchthat (jsc.string, s => {
       try {
-        return new RegExp (s, flags);
-      } catch (err) {
-        return /(?:)/;
```
davidchambers, line 1661: Is `/(?:)/` no longer necessary because `jsc.suchthat` rejects strings that are not regular expressions?

Avaq: Yes.

test/index.js
```diff
@@ -23,24 +24,23 @@ const {withUnstableArraySort} = require ('./quicksort');
...
-    let list = Nil;
-    for (let idx = 0; idx < xs.length; idx += 1) list = Cons (xs[idx], list);
-    return list;
-  })
-);
+const ListArb = arb => smapArb (arrayToList) (listToArray) (jsc.array (arb));
 
 //  MaybeArb :: Arbitrary a -> Arbitrary (Maybe a)
-const MaybeArb = arb => jsc.oneof (jsc.constant (Nothing), arb.smap (Just));
+const MaybeArb = arb => jsc.oneof (jsc.constant (Nothing), mapArb (Just) (arb));
```
Avaq, line 43: It might be worth adding a custom shrink function here to get back to an `Arb a` given an `Arb (Maybe a)`. :thinking:

Avaq: Hm. No. :P

Avaq:
@davidchambers Why did the tests fail?

test/index.js
```diff
@@ -77,6 +77,13 @@ function args() {
   return arguments;
 }
 
+//  arrayToList :: Array a -> List a
+function arrayToList(xs) {
+  let list = Nil;
+  for (let idx = 0; idx < xs.length; idx += 1) list = Cons (xs[idx], list);
+  return list;
+};
```
davidchambers, line 85: Let's use an arrow function.

We use function declarations below to enable us to verify the correctness of implementations by checking `arguments.length` (`this.map (f)` is not an acceptable implementation of `Array#fantasy-land/map` due to the extra arguments). Such assertions are not required here, so we can use an arrow function. :)

davidchambers (approved):
:clap:

PR #163, opened 2021-08-19
Principled Setoid and Ord with improved performance

Avaq:
Opened a draft PR for discussion.
See https://github.com/sanctuary-js/sanctuary-type-classes/pull/151#discussion_r692184710.

Remember we did that thing where we made `Z.equals` more useful by having it fall back to structural equality for garbage inputs (#154)? We ended up with a useful, but unprincipled and slow, equality function. That's fine and all, but I don't like that:

- Sanctuary is now forced to use this slow function, despite already having ensured that inputs are Setoids of the same type.
- Many Setoid implementations build on `Z.equals`, making them slow for no reason.
- Some Ord implementations, and `Z.lte` itself, builds on Z.equals.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1245,22 +1245,14 @@
     },
   }]);
 
-  //# equals :: (a, b) -> Boolean
+  //# equals :: Setoid a => (a, a) -> Boolean
   //.
   //. Returns `true` if its arguments are equal; `false` otherwise.
   //.
-  //. Specifically:
-  //.
-  //.   - Arguments with different [type identities][] are unequal.
-  //.
-  //.   - If the first argument has a [`fantasy-land/equals`][] method,
-  //.     that method is invoked to determine whether the arguments are
-  //.     equal (`fantasy-land/equals` implementations are provided for the
-  //.     following built-in types: Null, Undefined, Boolean, Number, Date,
-  //.     RegExp, String, Array, Arguments, Error, Object, and Function).
-  //.
-  //.   - Otherwise, the arguments are equal if their
-  //.     [entries][`Object.entries`] are equal (according to this algorithm).
+  //. > [!IMPORTANT]
+  //. >
+  //. > Arguments must be Setoids of the same type; providing arguments of
+  //. > differing types, or non-Setoid arguments, has undefined behaviour.
   //.
   //. The algorithm supports circular data structures. Two arrays are equal
   //. if they have the same index paths and for each path have equal values.
@@ -1286,8 +1278,6 @@
     const $pairs = [];
 
     Z.equals = (x, y) => {
-      if (!(sameType (x, y))) return false;
-
       //  This algorithm for comparing circular data structures was
       //  suggested in <http://stackoverflow.com/a/40622794/312785>.
       if ($pairs.some (([xx, yy]) => xx === x && yy === y)) {
@@ -1296,9 +1286,7 @@
 
       $pairs.push ([x, y]);
       try {
-        return Z.Setoid.test (x) ?
-               Z.Setoid.methods.equals (y, x) :
-               Object$prototype$equals.call (x, y);
+        return Z.Setoid.methods.equals (y, x);
       } finally {
         $pairs.pop ();
       }
@@ -2385,7 +2373,6 @@
 //. [Semigroupoid]:             v:fantasyland/fantasy-land#semigroupoid
 //. [Setoid]:                   v:fantasyland/fantasy-land#setoid
 //. [Traversable]:              v:fantasyland/fantasy-land#traversable
-//. [`Object.entries`]:         https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries
 //. [`fantasy-land/alt`]:       v:fantasyland/fantasy-land#alt-method
 //. [`fantasy-land/ap`]:        v:fantasyland/fantasy-land#ap-method
 //. [`fantasy-land/bimap`]:     v:fantasyland/fantasy-land#bimap-method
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1286,12 +1288,25 @@
...
+      return Z.Setoid.test (x) ? Z.setoidEquals (x, y) : structEquals (x, y);
+    };
+  }
+
+  // setoidEquals :: Setoid a => (a, a) -> Boolean
+  {
+    //  $pairs :: Array (Array2 Any Any)
+    const $pairs = [];
 
+    Z.setoidEquals = (x, y) => {
```
Avaq, line 1306: I placed the function on the exports (without documentation) for two reasons:

- Consistency with other parts of the code.
- I'd like it to be a public function eventually; I believe Sanctuary could use it instead of the unprincipled equality function, since Sanctuary will have already checked that input types are Setoids of the same type all the way down.

Avaq:
The first commit is problematic. Currently Z.equals doesn't check that *both* inputs are the same Setoid all the way down; it checks only `x` before dispatching to `setoidEquals`, meaning it will dispatch with garbage inputs such as `equals ([null], [42])` for which `setoidEquals` will happily return `true` because of the Setoid implementation for `null`.

Avaq:
This probably means that we'd have to have two completely separate implementations for `equals` and `setoidEquals`, with two separate type classes. This is something I hoped to avoid initially.

index.js
davidchambers: I agree with exporting this function. ADT libraries could use `Z.setoidEquals` in their `fantasy-land/equals` implementations, while `Z.equals` would remain available for use in test suites.

index.js
```diff
@@ -1287,8 +1289,6 @@
     const $pairs = [];
 
     Z.equals = (x, y) => {
-      if (!(sameType (x, y))) return false;
```
Avaq, line 1290: The removal of this line; and ...

index.js
```diff
@@ -1245,22 +1247,12 @@
...
-  //.   - Arguments with different [type identities][] are unequal.
-  //.
-  //.   - If the first argument has a [`fantasy-land/equals`][] method,
-  //.     that method is invoked to determine whether the arguments are
-  //.     equal (`fantasy-land/equals` implementations are provided for the
-  //.     following built-in types: Null, Undefined, Boolean, Number, Date,
-  //.     RegExp, String, Array, Arguments, Error, Object, and Function).
-  //.
-  //.   - Otherwise, the arguments are equal if their
-  //.     [entries][`Object.entries`] are equal (according to this algorithm).
```
davidchambers, line 1263: Could you remove the unused link definitions? :)

Avaq:
Right, so I removed the whole structural deep equality function, which I believe we said (face to face) we should reimplement in a separate package if the need is great enough.

Avaq:
The effect of these changes on performance:

```
$ npm run bench -- --benchmark old-vs-new --match '*{equals,lte}*'

> sanctuary-type-classes@13.0.0 bench
> sanctuary-benchmark "--benchmark" "old-vs-new" "--match" "*{equals,lte}*"

Running benchmarks: old-vs-new...

┌─────────────────────────┬──────────────────────────┬──────────────────────────┬────────┬──────────┬───┐                                                                                                                                                      
│ suite                   │ old                      │ new                      │ diff   │ change   │ α │
├─────────────────────────┼──────────────────────────┼──────────────────────────┼────────┼──────────┼───┤
│ methods.equals.Identity │ 202,779 Hz ±0.22% (n 94) │ 209,578 Hz ±0.59% (n 95) │ 001.6% │ +003.4%  │   │
├─────────────────────────┼──────────────────────────┼──────────────────────────┼────────┼──────────┼───┤
│ methods.equals.Object   │ 57,084 Hz ±1.84% (n 87)  │ 395,393 Hz ±2.48% (n 83) │ 074.8% │ +592.7%  │ ✓ │
├─────────────────────────┼──────────────────────────┼──────────────────────────┼────────┼──────────┼───┤
│ methods.lte.Identity    │ 145,642 Hz ±0.68% (n 93) │ 139,259 Hz ±4.06% (n 78) │ 002.2% │ -004.4%  │   │
├─────────────────────────┼──────────────────────────┼──────────────────────────┼────────┼──────────┼───┤
│ methods.lte.Object      │ 37,530 Hz ±2.05% (n 89)  │ 444,342 Hz ±2.20% (n 89) │ 084.4% │ +1084.0% │ ✓ │
└─────────────────────────┴──────────────────────────┴──────────────────────────┴────────┴──────────┴───┘
```

Avaq:
If we go this route, this will naturally have to be released as a breaking change. I don't expect many people were using `lte` and friends on anything other than real Ords, but I expect a few users to be impacted by the change to `equals`.

Avaq:
Although.. it hasn't been that long ago that `equals` was producing real useless garbage output for non-Setoids. Maybe it's not too bad.

test/index.js
```diff
@@ -724,11 +706,9 @@ test ('equals', () => {
   eq (Z.equals (Math.sin, Math.cos), false);
   eq (Z.equals (Identity (Identity (Identity (0))), Identity (Identity (Identity (0)))), true);
   eq (Z.equals (Identity (Identity (Identity (0))), Identity (Identity (Identity (1)))), false);
-  eq (Z.equals (Useless, Useless), true);
```
Avaq, line 727: I removed this. But prior to https://github.com/sanctuary-js/sanctuary-type-classes/pull/154/files#diff-5bb8db779819ddef5956a5d9d5949c05ef7445237656ca37bf2f02720271440bL676-R703, you had it in the test suite to explicitly test that Useless and Useless are not equal. It seems that you were testing that garbage input _actually_ produce garbage output. Should we have tests like that? I think we can leave unspecified behaviour totally unspecified.

davidchambers (approved):
:tada: :tada: :tada:

PR #181, opened 2024-11-10
remove benchmarking

davidchambers:
@Avaq added benchmarking in #60. I understand the appeal of having benchmarking embedded in this repository, but I'm now convinced that it belongs elsewhere. Maintaining the many Sanctuary projects is significantly more difficult due to their interconnectedness; embedded benchmarking is one of the culprits.

As an example, I'm working on converting this project to ES modules. This is problematic because [sanctuary-benchmark][1] expects __bench/baseline.js__ to be a CommonJS module, and we cannot `require` an ES module from a CommonJS module.

[1]: https://github.com/sanctuary-js/sanctuary-benchmark/blob/v1.0.1/bin/sanctuary-benchmark#L87


Avaq (approved):
I get it

